PR #7, opened 2026-04-09
Adding cuda-quantum's cudaq-guide skill

sacpis:
Adding cuda-quantum's cudaq-guide skill

Blocked on https://github.com/NVIDIA/cuda-quantum/pull/4229 getting merged in the main repo.

PR #8, opened 2026-04-10
Add GitHub Actions workflow to sync skills from product repos

sayalinvidia:
Implements the automated sync pipeline (Step 5 of onboarding) that sparse-checkouts the skills directory from each registered product repo and mirrors them into this catalog. Runs twice daily on a cron schedule and supports manual dispatch.

Registered repos: cuOpt, TensorRT-LLM, nemotron-voice-agent, NeMo Gym.

### DIFF
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -105,6 +105,28 @@ jobs:
             echo "⚠ Model-Optimizer checkout empty or missing — skipping to preserve existing catalog"
           fi
 
+      # ── CUDA-Q ───────────────────────────────────────────────
+      - name: Checkout CUDA-Q
+        continue-on-error: true
+        uses: actions/checkout@v4
+        with:
+          repository: NVIDIA/cuda-quantum
+          ref: main
+          path: .tmp/cuda-quantum
+          token: ${{ secrets.SKILLS_SYNC_PAT }}
+          sparse-checkout: |
+            .claude/skills/
+
+      - name: Copy CUDA-Q skills into catalog
+        run: |
+          if [ -d ".tmp/cuda-quantum/.claude/skills" ] && [ -n "$(ls -A .tmp/cuda-quantum/.claude/skills)" ]; then
+            mkdir -p skills/CUDA-Q
+            rsync -a --delete .tmp/cuda-quantum/.claude/skills/ skills/CUDA-Q/
+            echo "- CUDA-Q" >> /tmp/synced-products.txt
+          else
+            echo "⚠ CUDA-Q checkout empty or missing — skipping to preserve existing catalog"
+          fi
+
       # ── Megatron-Core ──────────────────────────────────────
       - name: Checkout Megatron-LM
         continue-on-error: true
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -20,6 +20,7 @@ on:
 
 permissions:
   contents: write
+  pull-requests: write
 
 concurrency:
   group: sync-skills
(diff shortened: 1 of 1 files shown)


mosheabr (changes requested):
Good start on the sync workflow, Sayali — the sparse-checkout approach and idempotent commit logic are solid. A few things to address before this is ready to merge:

## Critical

1. **Cross-repo auth will fail for private repos** — `actions/checkout@v4` uses the default `GITHUB_TOKEN`, which only has access to `NVIDIA/skills`. If any of the product repos (`NVIDIA/cuopt`, `NVIDIA/TensorRT-LLM`, etc.) are private, the checkout steps will 403. You'll need a PAT or GitHub App token:
   ```yaml
   with:
     token: ${{ secrets.SKILLS_SYNC_PAT }}
   ```

2. **Data loss risk if a checkout fails** — Each product block does `rm -rf skills/<product>` *before* rsync. If the checkout step fails (repo moved, branch renamed, transient error), you've deleted the existing catalog copy with nothing to replace it. Fix: guard the `rm -rf` so it only runs when `.tmp/<product>/skills/` actually exists and is non-empty, or move the delete into a conditional.

3. **Missing NeMo Evaluator** — The catalog currently lists 5 products (cuOpt, TensorRT-LLM, Nemotron Voice Agent, NeMo Gym, **NeMo Evaluator**). The workflow only syncs 4 — NeMo Evaluator needs a block added.

## Important

4. **Direct push to `main` bypasses branch protection** — Consider using `peter-evans/create-pull-request@v6` to open a PR instead of pushing directly, so changes can be reviewed before landing.

5. **No fault isolation** — If one product checkout fails, the entire job fails and no other products get synced. Consider `continue-on-error: true` on each checkout step, or a matrix strategy per product.

6. **No concurrency control** — If a manual dispatch overlaps with a cron run, two pushes could race. Add:
   ```yaml
   concurrency:
     group: sync-skills
     cancel-in-progress: true
   ```

## Minor

7. **`rm -rf` + `rsync --delete` is redundant** — `rsync --delete` already handles file removals from source. The `rm -rf` + `mkdir -p` before it is unnecessary.

8. **Static commit message** — `"chore: sync skills from product repos"` doesn't indicate which products changed. Would be helpful to include a summary.

9. **No failure notification** — If the cron sync silently fails, nobody knows. Consider adding a Slack or email notification step on failure.

mosheabr (approved):
Great revision, Sayali. All 9 items from the first review are addressed: PAT auth, data loss guards, fault isolation, concurrency, PR-based commits, dynamic commit messages, and failure notifications. This is solid.

One thing to add before merging: **CUDA-Q** was just merged into the catalog (#7). The sync workflow needs a block for it:

```yaml
# -- CUDA-Q --
- name: Checkout CUDA-Q
  continue-on-error: true
  uses: actions/checkout@v4
  with:
    repository: NVIDIA/cuda-quantum
    ref: main
    path: .tmp/cuda-quantum
    token: ${{ secrets.SKILLS_SYNC_PAT }}
    sparse-checkout: |
      .claude/skills/

- name: Copy CUDA-Q skills into catalog
  run: |
    if [ -d ".tmp/cuda-quantum/.claude/skills" ] && [ -n "$(ls -A .tmp/cuda-quantum/.claude/skills)" ]; then
      mkdir -p skills/CUDA-Q
      rsync -a --delete .tmp/cuda-quantum/.claude/skills/ skills/CUDA-Q/
      echo "- CUDA-Q" >> /tmp/synced-products.txt
    else
      echo "⚠ CUDA-Q checkout empty or missing — skipping to preserve existing catalog"
    fi
```

Once that's added, this is ready to go.

mosheabr:
CUDA-Q block looks good. All 10 products covered, all review items addressed. This is ready to merge whenever you mark it ready for review.

sayalinvidia:
Thank you @mosheabr 
Added CUDA-Q (NVIDIA/cuda-quantum) to the sync workflow as well!

PR #17, opened 2026-04-15
Add NVIDIA RAG Blueprint to skills catalog

kumar-punit:
Add the RAG Blueprint skill to the Available Skills table and Getting Help & Contributing section. The skill enables AI agents to deploy, configure, troubleshoot, and manage the NVIDIA RAG pipeline via Docker Compose or Helm.


mosheabr:
LGTM @kumar-punit — verified against the source repo:

- Skill count (1) is correct — single `rag-blueprint` skill directory
- Skill structure is solid — `SKILL.md` + `references/` with deploy, configure, troubleshoot, and shutdown guides (~30KB of reference material)
- Hosted under `NVIDIA-AI-Blueprints` org — consistent with existing catalog entries
- Contributing/Issues/Discussions/Security links all correct

One heads up: PR #19 is about to land which moves the catalog to a config-driven `products.yml`. After that merges you'll want to add a `products.yml` entry for RAG Blueprint so the sync workflow picks it up automatically. But that can be a follow-up.

Approving — merging now.

PR #24, opened 2026-04-17
Track and report components that fail to sync

sayalinvidia:
## Summary

- Log failed components with reason (clone failed or path empty/missing)
- Include failed components in the sync PR body so reviewers see what was skipped
- Open a GitHub issue listing failed components when the overall sync succeeds but some were skipped
- Existing total-failure issue (when nothing syncs) is unchanged

## Example PR body with failures

**Components synced:**
- CUDA-Q
- cuOpt
- TensorRT-LLM

**Components that failed to sync:**
- **Megatron-Core** (NVIDIA/Megatron-LM): `.claude/skills/` empty or missing

## Test plan

- [x] Tested on [sayali-skills-test](https://github.com/sayalinvidia/sayali-skills-test) — Megatron-Core correctly reported as failed, issue created

PR #29, opened 2026-04-27
Report only changed components in sync PR title and body

sayalinvidia:
The rsync runs every component each cycle but only some have upstream changes; the PR previously claimed all rsynced components were "synced" regardless of diff. Track rsynced vs. changed separately so the title and body reflect actual content updates and the "fail if nothing synced" health check still works on quiet days.

### DIFF
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -42,7 +42,8 @@ jobs:
           GH_TOKEN: ${{ secrets.SKILLS_SYNC_PAT }}
         run: |
           set -euo pipefail
-          truncate -s 0 /tmp/synced-components.txt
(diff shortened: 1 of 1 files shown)


sayalinvidia:
Verified on the test repo:
*PR generated before and after the changes*
_Nemotron Voice Agent is correctly excluded because nothing in it changed upstream since the last sync_ 
image

mosheabr:
Reviewed — clean implementation of the rsynced/changed split. The early-exit when no content changes prevents spurious PRs without dropping failures (those still flow through the auto-issue path from #24). Merging.

Thanks @sayalinvidia.

mosheabr (approved):
Reviewed and approved. Clean rsynced/changed split, preserves the failure path from #24, no-op on quiet days. Ready to merge.

PR #32, opened 2026-04-28
Add per-component Version column to README

sayalinvidia:
## Onboarding type

- [ ] New product onboarding (`components.yml` entry)

- [ ] `components.yml` unchanged (catalog set is unaffected by this PR)
- [ ] README table renders correctly with the new Version column
- [ ] Sync workflow change reviewed end-to-end (capture + patch steps)
- [ ] No new license or third-party dependency

## Other context

Adds a **Version** column to the Available Skills table that auto-populates with each component's upstream commit SHA and date, giving consumers per-component version visibility without leaving the catalog.

### What changes

1. **README markers** — each row gets a cell wrapped in HTML-comment markers: `<!-- sync:slug:version -->...<!-- /sync:slug:version -->`. Initial values are em dashes; RAG Blueprint stays plain `—` since it isn't in `components.yml`.
2. **Sync workflow capture** — after each successful clone, records upstream short SHA, full SHA, committer date, and repo to `/tmp/sync-versions.txt`.
3. **Sync workflow patch** — new step sed-replaces each marker block with `` [`abc1234`](https://github.com/<repo>/commit/<full_sha>) · YYYY-MM-DD ``. Components without markers are silently skipped.

### Design choices

- One combined "Version" column rather than separate SHA/date — less table bloat, both pieces of info visible.
- Date is the **upstream committer date** (`%cs`), not the sync run date — reflects when the version was created upstream, so readers can judge freshness ("3 weeks old").
- Slug derivation: `name | tr 'A-Z ' 'a-z-'` (e.g. `Nemotron Voice Agent` → `nemotron-voice-agent`).
- Sed delimiter `|` because URLs contain slashes; cell content has no pipes.
- Patch step runs **after** "Fail if nothing synced" so total failures don't dirty the README.
- New components onboarded later need to add the marker block to their README row; the workflow won't insert markers automatically.

### Known limitation

The committer date uses the committer's local timezone (`%cs`), so a commit at 22:44 UTC by a `+05:30` committer shows as the next day. Easy follow-up if cross-component date comparison matters: switch to `TZ=UTC git show -s --date=short --format=%cd`.


sayalinvidia:
verified updates on test repo:
image

mosheabr (approved):
Approving — implementation is solid, design choices well-justified (single Version column, upstream committer date, sed-with-pipe-delimiter, ordering after "Fail if nothing synced"). Verified slug derivation matches all 10 markers, sed pattern is safe, and idempotency works.

Before merge: can you rebase with `git rebase --signoff origin/main && git push -f` to add DCO sign-offs to the three commits? Lets us land this without `--admin` override and serves as the dry-run for the DCO enforcement work you're prepping.

Side observation: this PR surfaced that RAG Blueprint is in the README but not in components.yml (PR #17 added the README row but skipped the components.yml entry, so skills/rag/ doesn't exist). Tracking separately — not your PR's problem to fix. You correctly kept its Version cell as plain `—`.

PR #83, opened 2026-05-23
Enforce skill.oms.sig and skill-card.md in sync pipeline

sayalinvidia:
## Summary
- Skills missing either `skill.oms.sig` or `skill-card.md` are now **dropped from the catalog** before the sync PR is opened — compliance is no longer informational/opt-in.
- The check runs **per-skill**, not per-component: e.g. if cuopt has 5 skills and only 1 is compliant, that one syncs and the other 4 are dropped.
- Dropped skills surface in two places: a "Skills dropped" section in every sync PR body, and a single aggregated tracking issue labeled `missing-compliance` (auto-updates each sync, auto-closes when everything is compliant).

## What changed
- Replaced the previous informational "Scan for missing skill signatures" step with a mandatory "Enforce skill compliance (signature + card)" step that `rm -rf`s non-compliant skill directories ([sync-skills.yml:141-178](.github/workflows/sync-skills.yml#L141-L178)).
- Split issue tracking into its own step under a new `missing-compliance` label, replacing the older `missing-signatures` label ([sync-skills.yml:180-237](.github/workflows/sync-skills.yml#L180-L237)).
- Updated the PR summary so a dropped skill alone is enough to trigger a sync PR (catches the case where a previously-synced skill goes non-compliant upstream).

## Test plan
- [ ] Manually trigger the workflow via `workflow_dispatch` and confirm:
  - [ ] Compliant skills (e.g. `cuopt/cuopt-numerical-optimization-api-cli` — has both files) sync through normally
  - [ ] Non-compliant skills are removed from the catalog and listed in the PR body
  - [ ] Tracking issue is created/updated with the dropped skills grouped by product
  - [ ] Tracking issue auto-closes if all skills become compliant
- [ ] Confirm the workflow YAML parses cleanly (already verified locally)

### DIFF
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -138,7 +138,45 @@ jobs:
             exit 1
           fi
 
-      - name: Scan for missing skill signatures
+      - name: Enforce skill compliance (signature + card)
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          set -euo pipefail
+
+          # Every catalog skill (defined by SKILL.md) must ship with:
+          #   - skill.oms.sig  (detached NVIDIA signature)
+          #   - skill-card.md  (skill metadata card)
+          # Skills missing either file are dropped from the catalog before
+          # the PR is created — non-compliant skills do not get merged.
+          dropped=/tmp/dropped-skills.txt
+          truncate -s 0 "$dropped"
+
+          while IFS= read -r skill_md; do
+            dir=$(dirname "$skill_md")
+            missing=""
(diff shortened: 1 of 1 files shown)


sayalinvidia:
Note: with this change, any skill missing either skill.oms.sig or skill-card.md will be dropped from the catalog before the sync PR is opened 
Verified changes on the test repo! 
image

sayalinvidia:
Verified in the test repo! Currently, only cu-opt has (some) not all skills signed along with a skill card! 
Unverified skills from cu-opt will also be dropped!
image

mosheabr (approved):
Approved per catalog-pr-reviewer checklist.

Scope is correct: one workflow file (+79/-29), per-skill compliance enforcement, orphan-product cleanup, tracking issue under `missing-compliance` label. DCO + verify-authors both green. The diff also auto-marks products with dropped skills as 'changed' so the sync PR opens even when the only diff is a removal — good catch on the previously-compliant-now-non-compliant case.

Aligns with the 2026-05-27 signing-baseline cutover we communicated to product teams.

LGTM — over to you on when to merge.

PR #99, opened 2026-05-27
Register AIQ skills

AjayThorve:
## Summary
- Register AIQ as a component in the skills catalog.
- Add catalog mappings for aiq-research and aiq-deploy.
- Use the default source branch by omitting an explicit ref override.

## Validation
- Component file is named components.d/aiq.yml and uses name: AIQ so the name-derived slug is aiq.
- Commit includes DCO sign-off.

PR #102, opened 2026-05-27
Use AIQ develop branch for sync

AjayThorve:
## Summary
- Set the AIQ component sync ref to develop.

## Why
- The AIQ repository default branch is develop, while the catalog sync workflow falls back to main when ref is omitted.
- Without this ref, the sync workflow cannot find skills/aiq-research or skills/aiq-deploy, so AIQ does not get copied into the catalog or listed in README.

## Validation
- Verified NVIDIA-AI-Blueprints/aiq default branch is develop.
- Commit includes DCO sign-off.

### DIFF
--- a/components.d/aiq.yml
+++ b/components.d/aiq.yml
@@ -1,5 +1,6 @@
 name: AIQ
 repo: NVIDIA-AI-Blueprints/aiq
+ref: develop
 description: NVIDIA AI-Q Blueprint - deploy local AI-Q services and run shallow or deep research workflows as agent skills.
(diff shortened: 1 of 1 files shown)


mosheabr (approved):
Approving — one-line fix that correctly diagnoses the sync miss on PR #99/100. NVIDIA-AI-Blueprints/aiq default branch is develop, the sync defaulted to main, so the sparse-checkout came back empty. `ref: develop` resolves it.

Merging and re-triggering the sync.

PR #152, opened 2026-05-29
fix(nvidia-skills): correct include_skills paths after catalog flatten

bryan-anthropic:
### Summary

The `nvidia-skills` plugin currently builds with an empty `skills/` tree, so it installs without any skills.

`plugins.d/nvidia-skills.yml` curates two cuOpt skills via `skills/cuopt/<skill>/`, but the catalog was flattened — those skills now live directly at `skills/cuopt-routing-api-python/` and `skills/cuopt-user-rules/` (no `cuopt/` parent directory). Since the configured paths no longer resolve, `expand_skill_paths()` in `build-plugins.py` soft-skips both entries (by design, so curation drift doesn't block the daily sync), and the build regenerates `plugins/nvidia-skills/` with an empty `skills/` tree.

This is the `plugins.d/` companion to #122 ("components.d/cuopt: flat layout migration + remove orphan `skills/cuopt/`"). That PR migrated the cuOpt catalog to the flat layout and removed `skills/cuopt/`; this PR updates the one remaining reference to the old path — the `nvidia-skills` plugin curation.

### Fix

Point the two `include_skills` entries at the current catalog layout and regenerate the bundle:

```
.github/scripts/build-plugins.py --only nvidia-skills
```

This is the only source change; the `plugins/nvidia-skills/skills/` tree in this PR is the regenerated build output.

### Verification

- `.github/scripts/build-plugins.py --only nvidia-skills --check` → `✓ no drift`
- `claude plugin validate plugins/nvidia-skills` → `✔ Validation passed`
- Rebuilt bundle contains both skills (`cuopt-routing-api-python`, `cuopt-user-rules`) with `SKILL.md`, references, evals, and `skill.oms.sig` signatures intact.


mosheabr:
Thanks for catching this Bryan — really appreciate the quick fix on the include_skills paths after the catalog flatten.

Closing as superseded. The plugin curation has been updated internally since (`plugins.d/nvidia-skills.yml` now curates 11 skills, all on the flat-layout paths) and `plugins/nvidia-skills/skills/` is fully populated on main. The empty-tree symptom is resolved without merging this PR. Note that the curation evolved past your fix — `cuopt-routing-api-python` got dropped from the curated set entirely, and `cuopt-user-rules` is now in.

Thanks again for flagging — and for keeping an eye on the build output. Ping me anytime you spot drift like this.

PR #246, opened 2026-06-06
Metadata automation

jim-nvidia:
## Onboarding type

## For new product onboarding — author affirmations

By submitting this PR, I confirm on behalf of my team:

- [ ] **Skills cleared for open source release** per NVIDIA's internal IP review process (six-question check, all answers affirmative)
- [ ] **License selected:** Apache 2.0 / CC-BY 4.0 / Dual (Apache 2.0 + CC-BY 4.0). Specify: _____
- [ ] **No new license or new third-party component** introduced beyond what the source repo already carries
- [ ] **Source repo is public and under an NVIDIA-owned GitHub org**
- [ ] `.agents/skills/` or `skills/` path used for new entries (or existing path retained for legacy entries per `components.d/<slug>.yml`)

> NVIDIA contributors: see the internal onboarding guide for the IP review process details and license selection.

- [ ] Author confirmations above are checked
- [ ] `components.d/<slug>.yml` entry valid (required fields, unique `catalog_dir`, path exists in source repo, filename slug matches name)
- [ ] `SKILL.md` frontmatter spec-compliant (at least one sampled)
- [ ] No new license or third-party dependency requiring OSRB filing

## All PRs

      If you forgot, run `git rebase --signoff origin/main && git push --force-with-lease` to retroactively sign all commits in your branch.

## Other context (for non-onboarding PRs)

cc: @sayalinvidia @jasonnvidia

**Engineering sync completed June 9, 2026.** Reviewed with Jason Dudash, Moshe Abramovitch, and Sayali Kandarkar. Ready for engineering review, testing, and merge.

Introduces an end-to-end generator for the catalog-wide `metadata.json` and `skills.sh.json` files, plus the schemas, configs, workflow, and docs that support it.

### What it does

**Generator** (`.github/scripts/marketplace/generate-skill-metadata.py`):
- Walks `skills/` and parses `SKILL.md` frontmatter
- Carries forward existing values from the prior `metadata.json` baseline for byte-stable regen runs (unchanged skills produce zero AI calls)
- For materially changed skills (name or description changed), runs an "amend" mode: passes existing classification values as context and asks the model per-field whether to keep or update — biased toward preservation
- Falls back to an OpenAI-compatible LLM enrichment call (NVIDIA Inference API by default) for required fields the deterministic path cannot resolve, with strict JSON contract and no enum invention
- Per-skill enrichment failures go to a `skill_warnings` list; valid skills write regardless (partial-success exit code 1)
- Transient API failures (rate limits, timeouts, server errors) are retried up to 3 times with exponential backoff (1s, 2s, 4s) before giving up
- When skills are omitted, a structured warnings file is written and surfaced prominently in the GitHub tracking issue — listing each skipped skill and its reason
- Validates every output against the schemas before writing

**Schemas and configs** (`.github/scripts/marketplace/`):
- `metadata.schema.json` — single source of truth for controlled vocabulary
- `skills-sh.schema.json` — output structure for `skills.sh.json`
- `skills-subdomains.json` — subdomain titles, descriptions, ordering (keys are in canonical curated sequence; generator preserves insertion order)
- `metadata-exclusions.yaml` — temporary withholding list (currently empty)

**Workflow** (`.github/workflows/generate-skill-metadata.yml`):
- pull_request: runs --check --no-ai on changes to the generator scripts, component configs, or output files; PRs must produce byte-stable output. Note: skills/** is intentionally excluded from this trigger — new skills arrive via the automated sync and are handled by the post-sync regenerate job.
- `workflow_dispatch` and post-sync: regenerates with AI; opens an auto-PR on changes; opens or updates a tracking issue on validation failure

**Docs** (`.github/scripts/marketplace/README.md`): local usage, CI behavior, AI contract.

### Setup required (repo admin)

Before the workflow can run against `NVIDIA/skills`, a repo admin must configure the following in **Settings → Secrets and variables → Actions**:

| Name | Type | Value |
|---|---|---|
| `INFERENCE_API_KEY` | Repository **secret** | NVIDIA Inference API key |
| `INFERENCE_MODEL` | Repository **variable** | Model name (e.g. `openai/gpt-5.5`) |
| `INFERENCE_API_URL` | Repository **variable** | Optional — defaults to NVIDIA Inference API endpoint |

### Test plan

- [x] Generator runs locally end-to-end against the full skills tree (gpt-5.5 via Inference API)
- [x] --check --no-ai mode produces byte-stable output on an unmodified tree
- [x] Amend mode tested: fill-mode and amend-mode calls verified against live API
- [x] Partial-success path: a skill failing AI enrichment does not block output for others
- [x] Group ordering: subdomain sequence in skills.sh.json matches skills-subdomains.json insertion order
- [x] CI checks passing (DCO, Verify Authors, Generate Skill Metadata / check)
- [ ] Engineering team validates end-to-end in NVIDIA/skills with secrets configured before enabling for production
- [ ] CI workflow run confirmed working in NVIDIA/skills (requires secrets configured by admin)


sayalinvidia:
## Review findings from test run

Tested this branch by merging into a test repo and triggering the `regenerate` workflow. Two bugs surfaced:

---

### Bug 1: `TypeError: Object of type date is not JSON serializable`

**File:** `.github/scripts/marketplace/generate-skill-metadata.py`, line 540

When a `SKILL.md` frontmatter contains a YAML date field (e.g. `reviewed: 2024-01-15`), PyYAML parses it as a `datetime.date` object. The AI client then tries to serialize the full `skill_frontmatter` dict via `json.dumps`, which crashes:

\`\`\`
TypeError: Object of type date is not JSON serializable
when serializing dict item 'reviewed'
when serializing dict item 'skill_frontmatter'
\`\`\`

**Fix — one-word change on line 540:**
\`\`\`python
# before
"content": json.dumps(user, ensure_ascii=False),

# after
"content": json.dumps(user, ensure_ascii=False, default=str),
\`\`\`

`default=str` safely coerces any non-serializable value (date, datetime, Path, etc.) to its string representation before sending to the API.

---

### Bug 2: Generator failure silently skips the tracking issue step

**File:** `.github/workflows/generate-skill-metadata.yml`, `Run generator` step

GitHub Actions runs steps with `bash -eo pipefail` by default. When the generator exits non-zero, `-e` fires immediately — the `echo "exit_code=${PIPESTATUS[0]}"` line never executes, so the `exit_code` output is never written. Because subsequent steps have no `if: always()` guard, **the "Open or update validation issue" step is silently skipped** — exactly the failure path it's meant to handle.

Confirmed live: the generator exited with code 1 and every downstream step showed `skipped`.

**Fix:**
\`\`\`yaml
- name: Run generator
  id: generate
  continue-on-error: true
  env:
    INFERENCE_API_KEY: ${{ secrets.INFERENCE_API_KEY }}
    INFERENCE_API_URL: ${{ vars.INFERENCE_API_URL }}
    INFERENCE_MODEL: ${{ vars.INFERENCE_MODEL }}
  run: |
    rc=0
    python3 .github/scripts/marketplace/generate-skill-metadata.py 2>&1 | tee /tmp/generator.log || rc=$?
    echo "exit_code=${rc}" >> "$GITHUB_OUTPUT"
    exit ${rc}
\`\`\`

`continue-on-error: true` lets subsequent steps run regardless of the generator's exit code. Capturing `rc` before `exit` ensures the output is always written.

---

cc @jim-nvidia — both bugs are small fixes but bug 2 means the tracking issue mechanism won't fire at all on any validation failure until it's patched.

sayalinvidia:
Thanks for the updates @jim-nvidia — the changes address the prior round's feedback well. 
Ran the generator against a test repo with the API key wired up; it completed end-to-end successfully. 

**A few remaining items below.**

**Medium — unhandled GeneratorError in main()**

If discover_skills hits a SKILL.md with missing frontmatter, the GeneratorError it raises propagates uncaught all the way to sys.exit(main()) — the user gets a Python traceback instead of a clean error message and exit code 1. Other GeneratorError paths are caught. Suggest wrapping the discover_skills call in main() with a try/except GeneratorError:


try:
 ```
   skills = discover_skills(skills_root, exclusions)
except GeneratorError as e:
    print(f"error: {e}", file=sys.stderr)
    return 1
Medium — retry loop doesn't catch all RequestException subtypes
```

The _AIClient.__call__ retry loop catches requests.exceptions.Timeout and requests.exceptions.ConnectionError specifically. Other RequestException subclasses (e.g. TooManyRedirects, InvalidURL) would propagate as an unhandled exception rather than a clean EnrichmentError. Suggest catching requests.exceptions.RequestException as the base class and narrowing from there.

**Medium — rename detection matches on name only**

`classify() detects renames via old_name in name_to_added. The inline comment says "name+description" but only name is compared. Two unrelated skills with the same name would silently inherit each other's metadata. Worth either matching on both fields or updating the comment to reflect the actual behavior.`

**From live test — duplicate path error message loses context**

When validate_skills_sh_uniqueness fires, the message reads:

skill deepstream-dev appears in both Vision AI and Vision AI

Same group name on both sides makes the error hard to diagnose — you can't tell which two paths are colliding. Showing the conflicting paths would be more actionable:

skill deepstream-dev appears at both skills/deepstream-dev and skills/deepstream/deepstream-dev — both map to Vision AI

_(This surfaced in testing because the test repo has stale nested-layout directories alongside the flat layout — not a real-world issue for the main catalog, but a nice UX improvement.)_

sayalinvidia (approved):
Thanks for pushing all the fixes @jim-nvidia 
LGTM - merging!

PR #247, opened 2026-06-06
Add Super Orchestrator skill

nikhil49023:
This PR registers the Super Orchestrator skill (https://github.com/nikhil49023/super-orchestrator-skill) in the components index. It has been validated with SkillSpector and scored 0/100 (Safe).

### DIFF
--- a/components.d/super-orchestrator.yml
+++ b/components.d/super-orchestrator.yml
@@ -1,0 +1,6 @@
+name: Super Orchestrator
+repo: nikhil49023/super-orchestrator-skill
+description: Master orchestration skill utilizing code-graph-first, local firecrawl, dynamic subagents in isolated workspaces, automated token logging, and free opencode models for lightweight tasks.
+skills:
+  - path: .
+    catalog_dir: super-orchestrator


mosheabr:
Hi Nikhil, thanks for putting in the effort here, and for running it through SkillSpector before submitting.

Couple of things blocking this from landing in the catalog:

1. **Source repo location.** Catalog entries have to point at a public repo under an NVIDIA-owned GitHub org (`NVIDIA/*`, `NVIDIA-NeMo/*`, `NVIDIA-AI-Blueprints/*`, etc.). `nikhil49023/super-orchestrator-skill` is a personal repo, so it doesn't qualify under the current rules.
2. **External contributor.** The catalog is currently restricted to NVIDIA-authored skills, with CI enforcement on commit author affiliation. External skills need an OSRB-equivalent review path that doesn't exist yet for community submissions.

The work isn't lost — you can keep `super-orchestrator-skill` published on your own GitHub and users can install it directly from there. We're aware there's demand for a community-contribution path and it's something we're thinking about, but it's not in place today.

Closing this PR. Thanks again for the submission, and feel free to keep an eye on the repo for when external contribution opens up.

PR #259, opened 2026-06-09
feat(sync): resolve source-team contact for dropped/reverted skills

### LINKED ISSUE #251: feat(sync): tag source-team contact when sync drops or reverts a skill
## Problem

When the sync workflow drops or reverts a skill (missing artifacts, sig drift, or sig missing), only the catalog PIC is notified via the rolling tracker issue. The actual source team that owns the skill isn't tagged or pinged. Today this means catalog PIC has to manually Slack each team after every sync — exactly what we've been doing this morning across VSS, AutoModel, cuOpt, NemoClaw, Physical AI, etc.

Follow-up on PR #235 (sig drift + missing-sig defense).

## Proposal

Auto-resolve the source-team contact for each dropped/held skill from the `components.d/<slug>.yml` git history. The author of the original "add" commit for that YML is the canonical contact.

### Resolution flow

For each skill in `/tmp/dropped-skills.txt`:

1. Find the owning YML via `grep -l "catalog_dir: $skill_dir" components.d/*.yml`
2. Get the author of the first commit that introduced that YML:
   ```bash
   git log --diff-filter=A --format='%ae' -- "$owner_yml" | tail -1
   ```
3. Resolve email to GitHub login (NVIDIA emails map deterministically; fall back to `gh api users/<login>` or `gh api search/users?q=in:email`)
4. Tag the resolved login in the tracker issue body and the sync PR body

### Optional `contacts:` field in YML

Add an optional schema field for teams who want to override the default (e.g., original onboarder left the company, or a team alias is preferred):

```yaml
name: VSS
repo: NVIDIA-AI-Blueprints/video-search-and-summarization
contacts:
  - "@zacharyhuang-nv"
  - "@hugomontanari"
skills:
  - ...
```

If `contacts:` is present, use it. Otherwise fall back to git-blame resolution.

### Output surface

Update the existing tracker-issue body partitioning to include a "Contact" column or trailer per component:

> **Skills held — signature drift:**
> - `skills/vss-deploy-dense-captioning` — reverted (4 file(s) changed without skill.oms.sig refresh). cc: @zacharyhuang-nv @hugomontanari

Same treatment in the sync PR body.

## Why this matters

- Closes the manual-Slack-after-every-sync loop
- Source teams find out immediately when their skill is held, instead of via a downstream catalog PIC ping a day later
- Keeps the catalog PIC's signal-to-noise high — they stop being the routing layer for every sig drift event

## Out of scope (future work)

- Opening issues directly on source product repos (requires PAT scope to write issues across NVIDIA orgs)
- Email/Slack channel routing (requires integration with a notification service)

## Implementation notes

- New step or in-line additions to existing drift / compliance / dropped-skills steps in `.github/workflows/sync-skills.yml`
- Cache email → GitHub login lookups within a single run to avoid rate limits on multi-skill drops
- Test fixtures: drift on a skill belonging to a known YML; verify the resolved handle matches the YML's `--diff-filter=A` author

## Priority

P1 — biggest workflow win after PR #235 itself. Without it, the sig drift gate fires correctly but the recovery loop stays manual.

mosheabr:
Closes #251.

## Summary

Adds a `Resolve skill contacts` step to `sync-skills.yml` that auto-tags the original onboarder of each dropped or reverted skill in the tracker issue body and sync PR body. Closes the manual-Slack loop where catalog PIC has to chase down the right team contact after every sync.

Implementation follows @sayalinvidia's design review on #251:
- **Drop the optional `contacts:` YML schema field** — reactive override can land later if first-onboarder data goes stale.
- **Git-blame-only resolution** — for each entry in `/tmp/dropped-skills.txt`, find the owning `components.d/<slug>.yml`, then `git log --diff-filter=A` to get the original onboarder's email, then resolve to GitHub login via `search/users`.
- **Loud workflow warning on resolution failure** — no silent skip when the email→login lookup returns empty.
- **Output to both surfaces** — tracker issue body and sync PR body, via a shared `lookup_contact` / `emit_with_contact` helper.
- **Dedicated step** — separate from drift/compliance for easier debug and independent skip.

## Behavior

For each dropped skill, the body lines now look like:

```
- `skills/dali-dynamic-mode` — reverted (2 file(s) changed without skill.oms.sig refresh) (cc: @michalzient)
```

When the resolver can't find a GitHub login for the onboarder email, the entry still appears without a `cc:` trailer and a `::warning::` is emitted in the workflow log:

```
::warning::could not resolve GitHub login for zacw@nvidia.com (onboarder of components.d/video-search-and-summarization.yml) — vss-summarize-video will land without contact tag
```

## File-backed cache

Resolution is cached in `/tmp/email-login-cache.txt` (one entry per unique email per run) so large drop batches make at most one API call per onboarder. Cache uses a `-` sentinel for tried-but-empty results so we don't retry within a run. File-backed rather than `declare -A` so the step works on any bash version.

## Verified locally

Synthetic `dropped-skills.txt` with 2 entries (vss-summarize-video, cuopt-install), run against the live catalog:

```
::warning::could not resolve GitHub login for zacw@nvidia.com (onboarder of components.d/video-search-and-summarization.yml) — vss-summarize-video will land without contact tag
  cuopt-install → @sayalinvidia (from skandarkar@nvidia.com via cuopt.yml)
Resolved 1/2 contacts.
```

## Onboarding type

- [ ] New product onboarding
- [x] Other (sync infrastructure)

## All PRs

- [x] DCO sign-off

### DIFF
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -354,6 +354,109 @@ jobs:
             done < "$dropped"
           fi
 
+      - name: Resolve skill contacts
+        # For each dropped/reverted skill, resolve the original onboarder
+        # via git blame on the owning components.d/<slug>.yml and emit a
+        # mapping (catalog_dir → @github-login) into /tmp/skill-contacts.txt.
+        # Downstream steps (tracker issue body, sync PR body) read this
+        # file and @-mention the contact so the source-team owner is
+        # notified directly instead of via a manual catalog-PIC ping.
+        #
+        # Resolution rules:
+        #   - YML owner: first file matching `catalog_dir: <dir>` under
+        #     components.d/*.yml (orphans without a YML are skipped with
+        #     a warning).
+        #   - Author email: `git log --diff-filter=A` on the YML — the
+        #     person who added the component is the canonical contact.
+        #   - Email → GH login: GitHub user search by email. The repo's
+        #     Verify Authors gate guarantees NVIDIA-issued emails, so the
+        #     lookup usually resolves. If it doesn't, we emit a loud
+        #     workflow warning rather than silently skipping the tag.
+        # Lookups are cached in-memory across the run to keep API calls
+        # well under the rate limit even on large drop batches.
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+
+          dropped=/tmp/dropped-skills.txt
+          contacts=/tmp/skill-contacts.txt
+          : > "$contacts"
+
+          if [ -s "$dropped" ]; then
+            :
+          else
+            echo "  (no dropped skills — nothing to resolve)"
+            exit 0
+          fi
+
+          # File-backed cache for email → GH login lookups so we make at
+          # most one API call per unique email even on large drop batches.
+          # File-backed rather than associative array so the step works
+          # under any bash version.
+          cache=/tmp/email-login-cache.txt
+          : > "$cache"
+
+          while IFS='|' read -r product dir reason; do
+            [ -z "$dir" ] && continue
+            cat_dir=$(basename "$dir")
+
+            # Find owning YML. Multiple `catalog_dir:` matches across
+            # components.d/ are not expected (uniqueness gate at PR time)
+            # but if it happens we take the first deterministically.
+            owner_yml=$(grep -l "catalog_dir: $cat_dir" components.d/*.yml 2>/dev/null | sort | head -1)
+            if [ -z "$owner_yml" ]; then
+              echo "::warning::no YML owner found for $dir — cannot resolve contact (orphan or schema mismatch)"
+              continue
+            fi
+
+            # Author of the first commit that added this YML = onboarder.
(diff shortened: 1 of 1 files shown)


sayalinvidia:
Thanks @mosheabr, Solid implementation:  minimal footprint, handles failure gracefully, matches the design exactly. 

_One correctness bug worth looking into:_

catalog_dir grep can false-positive. 
`The lookup grep -l "catalog_dir: $cat_dir" matches any substring — if a skill dir is named rag, it would also match catalog_dir: rag-blueprint. Should anchor it: grep -l "catalog_dir: ${cat_dir}$" (or grep -lP with word boundary).`

LGTM after resolving the issue mentioned above!

mosheabr:
Thanks @sayalinvidia — good catch, that was a real correctness bug. Pushed the anchor fix in e6fb8d7.

Verified locally: \`grep -l \"catalog_dir: rag\$\"\` no longer matches \`rag-blueprint\`, while \`grep -l \"catalog_dir: rag-blueprint\$\"\` still resolves correctly. Used \`\$\` end-anchor rather than \`-P\` word-boundary since YAML lines for components.d/*.yml don't carry trailing comments or whitespace, so \`\$\` is reliable and avoids the Perl dependency.

Ready for re-review when you have a minute.

PR #264, opened 2026-06-12
Clarify recommended skill directory path

holgerroth:
## Summary
- Clarifies that product repos should use `skills/` as the source-of-truth location for published skills.
- Clarifies that `.agents/skills/` is an agent runtime discovery/install location, except for repos intentionally structured as skills packs.
- Keeps the contribution guidance focused.

## Validation
- `git diff --check`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,12 +11,13 @@ For changes to the catalog itself (fixing links, adding a new product listing), 
 
 ## Recommended Skill Directory Path
 
-When publishing skills in your product repo, use one of these canonical paths so agents can discover them consistently:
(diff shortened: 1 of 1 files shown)


mosheabr (approved):
Thanks, Holger — this is a good correction. It aligns CONTRIBUTING.md with the canonical guidance (skills/ as the source of truth, `.agents/skills/` as an optional install/runtime symlink target), and the install-time copy/symlink framing is exactly right. Approving.

Minor follow-up (not blocking this PR): the edited block still references `components.yml`, which has been retired in favor of per-product `components.d/<slug>.yml`. I'll fix that in a quick follow-up PR so it doesn't hold this up.

PR #266, opened 2026-06-12
ci: verify skill content matches its signed manifest

mosheabr:
## Problem

Our checks confirm a skill's signature is **present** and **genuine**, but nothing confirms the files on `main` still **match** what that signature signed. A signature signs a manifest of per-file sha256 hashes — so a skill can carry a 100% genuine signature while its content has drifted (a team signs an early commit, keeps editing, and the sync picks up later content against the stale signature). "Signed" is not the same as "the signature is telling the truth about what's inside."

This gap is real and live: a full-catalog audit found **6 skills** whose files don't match their own signature, none of which any existing check caught.

## What this adds

A content-integrity check (`.github/scripts/verify_content_integrity.py` + workflow):
- Decodes each `skills/<skill>/skill.oms.sig` manifest, recomputes the sha256 of every signed file on disk, and compares.
- **Fails on any hash mismatch or missing signed file.**
- **pull_request:** checks only the skills changed in the PR — catches drift being *introduced*, without blocking on unrelated pre-existing state.
- **schedule (daily) + workflow_dispatch:** checks the whole catalog — safety net for drift already on `main`.
- Pure Python stdlib, runs against the local checkout: fast (local hashing, scoped to changed files), no network, no secrets, no extra deps.

This is complementary to the existing signature presence/authenticity gates — it adds the *freshness* dimension ("does the content still match the signature").

## Validation

Run against the current catalog it reports **195 ok / 6 drifted** — exactly the known set (VSS ×3, NemoClaw ×2, TAO ×1), which the owning teams are already remediating. PR-mode correctly scopes to changed skills; a clean diff passes.

## All PRs
- [x] All commits signed off with DCO.

### DIFF
--- a/.github/scripts/verify_content_integrity.py
+++ b/.github/scripts/verify_content_integrity.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2026 NVIDIA Corporation. All rights reserved.
+"""Verify that each skill's files still match its signed manifest.
+
+A ``skill.oms.sig`` is a Sigstore bundle wrapping a DSSE / in-toto statement
+whose ``predicate.resources`` lists every signed file together with its
+sha256 digest. Signature *presence* and *authenticity* tell you a genuine
+signature exists; they do NOT tell you the files still match what was signed.
+
+A skill can therefore carry a perfectly genuine signature while its content
+has drifted — e.g. a team signs commit 1, makes more commits, and the sync
+picks up a later content with the stale signature. This script closes that
+gap: for each signed file it recomputes the sha256 on disk and compares it to
+the signed digest, failing on any mismatch or missing file.
+
+Scope:
+  * pull_request  -> only the skills changed in the PR (fast; catches drift
+                     being introduced).
+  * schedule / workflow_dispatch / anything else -> the whole catalog
+                     (safety net; catches pre-existing drift already on main).
+
+Exit code 0 = all checked skills match their signatures; 1 = drift found.
+"""
+from __future__ import annotations
+
+import base64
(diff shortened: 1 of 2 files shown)


mosheabr:
@sayalinvidia — this is the content-integrity check from our thread. Flagging for your review since it's sync-pipeline-adjacent.

**What it does:** decodes each `skills/<skill>/skill.oms.sig` manifest, recomputes the sha256 of every signed file, and compares — failing on any mismatch or missing file. This is the *freshness* check (does the content still match the signature), complementary to the existing presence/authenticity gates.

**Scope:** on a PR it checks only the skills changed in that PR (catches drift being introduced, won't block on unrelated pre-existing state); a daily scheduled run + manual dispatch check the whole catalog.

**Validation:** run against the current catalog it reports 195 ok / 6 drifted — exactly the known set (VSS ×3, NemoClaw ×2, TAO ×1) those teams are already re-signing. PR-mode correctly scopes to changed skills; clean diffs pass. It's pure stdlib, runs on the local checkout (fast, no network, no secrets).

**One thing to note:** merging this makes the check *run + report*. To make a failure actually *block* merges, `content-integrity` needs to be added to `main`'s required status checks in branch protection (a one-time settings toggle) — same as how `dco`/`authors` are enforced today.

Happy to walk through it or adjust scope/behavior.

sayalinvidia (approved):
LGTM, looks like a solid, well-reasoned security addition!

A few small things worth a look:

signed_resources() accesses bundle["dsseEnvelope"]["payload"] and statement["predicate"]["resources"] with no key guards. Any deviation from the expected schema will throw a KeyError that gets swallowed by the outer except Exception — which works, but the resulting error message won't tell you much. 

Just flagging it; it's not a blocking issue!

PR #267, opened 2026-06-12
feat(sync): auto-ping source-repo owners when skills are dropped

sayalinvidia:
cc @mosheabr 

## Summary

Automates the manual pinging of source-repo teams when their skill updates are held back by the sync workflow (sig drift, missing artifacts). Closes the gap tracked in #216 — no more manually chasing teams to run `/nvskills-ci`.

## What changes

Three additions to `sync-skills.yml`:

**1. Source map** — during the rsync loop, every successfully rsynced skill records `catalog_dir|source_repo|source_path` to `/tmp/skill-source-map.txt` before the `.tmp` clones are deleted. This gives later steps the data they need without re-cloning.

**2. Snapshot step** (`Snapshot existing drop-tracker state`, before Track) — saves the current rolling issue body to a file so the ping step can tell which drops are new vs already tracked in a prior run.

**3. Ping step** (`Ping skill owners for new drops`, after Track) — for each *newly* dropped skill:
- Looks up source repo + path from the map
- Calls `gh api repos/<source_repo>/commits?path=<skill_path>&per_page=5` to resolve the last GitHub logins who touched that skill
- Posts **one aggregated comment** on the rolling tracker issue with @mentions for all affected owners

## Behavior

When `skills/aiq-research` drops for the first time, the comment looks like:

> **New skill drops this sync run — action required**
>
> - [`skills/aiq-research`](…) ([source: NVIDIA-AI-Blueprints/aiq](…)) — reverted (2 file(s) changed without skill.oms.sig refresh) — cc @alice @bob
>
> **How to fix:** run `/nvskills-ci` on a follow-up PR in the source repo…

- Teams are pinged **once per new drop**, not on every subsequent sync run while the skill stays broken
- @mentions are resolved from the GitHub commits API — no static contact list to maintain
- Degrades gracefully for private repos or unlinked accounts (comment still posts, just no @mention)
- Both new steps are `continue-on-error: true` so a ping failure never blocks the sync


sayalinvidia:
Closing in favor of #268 — rebased cleanly on current main, no conflicts.

PR #268, opened 2026-06-12
feat(sync): post comment pinging owners when skills newly drop

sayalinvidia:
cc @sayalinvidia

## Summary

`origin/main` already has a "Resolve skill contacts" step that resolves the `components.d/` owner for each dropped skill (via `git log --diff-filter=A` + email → GitHub login search) and surfaces `(cc: @login)` in the rolling tracker issue body. But **editing the issue body doesn't send GitHub notifications** — owners who aren't watching `NVIDIA/skills` never see it.

This PR adds the missing notification layer: when skills drop for the first time, post a single comment on the tracker issue with @mentions so owners get a direct GitHub notification.

## What changes

Two new steps in `sync-skills.yml` (no changes to existing steps):

**1. `Snapshot existing drop-tracker state`** (before Track) — saves the current issue body before it's overwritten, so the ping step can identify which drops are genuinely new this run.

**2. `Ping skill owners for new drops`** (after Track) — for each first-time drop:
- Reads the already-resolved `@login` from `/tmp/skill-contacts.txt` (populated by the existing "Resolve skill contacts" step — no duplicate API calls)
- Posts one aggregated comment on the tracker issue with all @mentions

## Behavior

- Owners are pinged **once per new drop**, not every 12-hour sync cycle while the skill stays broken
- Re-pinged if the skill gets fixed and breaks again
- Both steps are `continue-on-error: true` — a failure never blocks the sync
- Gracefully no-ops if contacts couldn't be resolved (comment still posts for skills that have no contact, just without a mention)

### DIFF
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -461,6 +461,19 @@ jobs:
             echo "Resolved $resolved/$total contacts."
           fi
 
+      - name: Snapshot existing drop-tracker state
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Save the pre-update issue body so "Ping skill owners" can
+          # determine which drops are new this run vs already tracked.
+          # Writes {} on any failure so downstream steps degrade gracefully.
+          gh issue list --label missing-compliance --state open \
+            --json number,body --jq '.[0] // {}' --limit 1 \
+            > /tmp/pre-update-tracker.json 2>/dev/null \
+            || echo '{}' > /tmp/pre-update-tracker.json
+
       - name: Track dropped skills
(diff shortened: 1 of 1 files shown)


mosheabr (approved):
Approved. Reviewed the bash logic end-to-end and the key correctness property holds: the dedup key (`` `$dir` `` backtick-quoted, full path) matches the format the Track step writes into the issue body, and `dropped-skills.txt` / `skill-contacts.txt` field layouts line up — so owners get pinged once per new drop, not every sync cycle. Sequencing (Snapshot → Track → Ping against the pre-update body) is right, first-run and all-compliant cases no-op cleanly, and both steps are `continue-on-error` so they can't block a sync. Nice — this closes the "issue body edit doesn't notify anyone" gap.

PR #300, opened 2026-06-29
chore(cuopt): update skill list in components.d

ramakrishnap-nv:
## Summary

- Remove `cuopt-developer`, `cuopt-user-rules`, `cuopt-server-common`, `cuopt-routing-formulation`, and `cuopt-skill-evolution` from the cuOpt skill list
- Add `cuopt-multi-objective-exploration`

## Test plan

- [ ] Verify `components.d/cuopt.yml` reflects the correct skill entries
- [ ] Confirm removed skills are no longer referenced
- [ ] Confirm `cuopt-multi-objective-exploration` path/catalog_dir are valid

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #305, opened 2026-06-30
feat(plugins): generate Cursor plugin alongside Claude and Codex

jasonnvidia:
## Summary

Adds a **Cursor** plugin track to the plugin build pipeline so each catalog plugin in `plugins.d/` is emitted for Cursor in addition to the existing Claude and Codex tracks — driven from the same single source of truth.

- `build-plugins.py`: new `render_cursor_plugin_json()` + `upsert_cursor_marketplace()`, wired into `build_catalog_plugin()` and `main()`, and added to the `--check` drift guard.
- Generates `plugins/<name>/.cursor-plugin/plugin.json` and a top-level `.cursor-plugin/marketplace.json`, parallel to `.claude-plugin/` and `.codex-plugin/` (+ their marketplaces).
- New `marketplace_enabled.cursor` opt-out flag, documented in `_defaults.yml` and `nvidia-skills.yml`.
- `validate-plugins.yml`: watches `.cursor-plugin/marketplace.json` and documents the Cursor validation story.
- Docs: `plugins.d/README.md` and the `build-plugins.py` docstring updated.

## Cursor-specific manifest decisions (per the [plugins reference](https://cursor.com/docs/reference/plugins))

- **Omits the `skills` field** so Cursor's folder-based discovery scans `skills/` (each `SKILL.md` subdir). Setting an explicit path would replace discovery and mis-treat the container as a single skill.
- Flat manifest (no Codex-style `interface{}` block) — Cursor's schema doesn't use it.
- Projects the author to `{name}` (Cursor's `author` is `name`/`email`); `marketplace.json` `owner` follows the documented `{name}` schema.
- Strips the leading `./` from `logo` so it resolves against `raw.githubusercontent.com`.

## Test plan

- [x] `build-plugins.sh` generates all three tracks; re-run is idempotent (`--check` clean).
- [x] Verified `.cursor-plugin/plugin.json` version parity with Claude/Codex (`1.0.3`).
- [x] Functional load test with `cursor-agent --plugin-dir <plugin>`: bundled skills (e.g. `nemoclaw-user-guide`) load and are usable via Cursor folder discovery. (Loading only fails when `--plugin-dir` points into this catalog repo itself, due to duplicate skill names colliding with the canonical `skills/` source — not an issue for real installs.)
- [x] Top-level marketplace `owner`/`metadata` preserved across rebuilds; only `plugins[]` regenerated.


jasonnvidia:
Folded the cuOpt bundle swap into this PR (separate commit, `chore(plugins): swap cuOpt skill in nvidia-skills bundle`).

- Replaced `skills/cuopt-user-rules/` (pruned from the cuOpt catalog in #300) with `skills/cuopt-numerical-optimization-api-python/` — the flagship cuOpt skill / README quickstart example — in `plugins.d/nvidia-skills.yml`.
- This removes the dangling-skill risk on the next sync and gives cuOpt a better representative across Claude, Codex, and Cursor.
- Structural change (skill set changed) auto-bumped the plugin `1.0.3 → 1.1.0` via `version-plugins.py`; all three manifests updated and the drift check is clean.

Thanks for the catch, @mosheabr — keeping it in one PR per your offer.

jasonnvidia:
Quick heads-up on the diff size: the bundle uses `skill_files: copy`, so swapping in `cuopt-numerical-optimization-api-python` pulled in all of its asset/benchmark/reference files too (~2.9k lines). Nothing weird going on — it's just how the copied skills ship — but wanted to call it out so the line count doesn't throw anyone during review. 🙂

mosheabr (approved):
Reviewed end to end. Cursor plugin track mirrors the existing Claude/Codex render+upsert functions exactly and is wired into the drift guard; cuOpt bundle swap removes the dangling-skill risk and correctly bumps 1.0.3 → 1.1.0 across all three manifests. DCO + author affiliation pass, swapped-in cuOpt skill carries sig/skill-card/evals/BENCHMARK, all four checks green. LGTM.

PR #326, opened 2026-07-06
Aggregate BENCHMARK.md data into machine-readable benchmarks.json

mosheabr:
## What

Aggregates every skill's `BENCHMARK.md` evaluation report into a single machine-readable `benchmarks.json` at the repo root, so downstream tooling and dashboards can consume per-skill benchmark scores and uplift data without parsing markdown.

## Contents

- `.github/scripts/aggregate_benchmarks.py` — walks `skills/*/BENCHMARK.md`, extracts the evaluation summary (date, profile, environment, tasks, threshold, verdict), and emits two flat tables: `skills[]` (per-skill summary with `has_results` flag and average uplift) and `results[]` (one row per skill × dimension × agent with `score_pct` + `uplift_pct`, denormalized with `catalog_dir` and `component` for joins). Agent columns are keyed off each results-table header, not position. Supports `--check` (exit 1 on drift).
- `benchmarks.json` — generated output (schema_version 2).
- `generate-skill-metadata.yml` — regeneration wired into the existing metadata workflow: PR mode validates `benchmarks.json` against sources; the post-sync/dispatch job regenerates it and includes it in the automated regeneration PR. Same freshness guarantees as `metadata.json`/`skills.sh.json`.

## Coverage

- 235 skills total; **212 with Tier-3 dimension results** (2,120 measurement rows), 23 listed in `skills_without_results` (their reports state Tier-3 rollup was not available).
- Catalog-wide average uplift across all dimensions/agents: **+25.6%**.

## Schema notes for consumers

- `catalog_dir` is the authoritative join key (one report has a placeholder in its `Skill:` field — `deepstream-generate-pipeline`).
- `uplift_pct` is null when a report had no baseline; consumers should treat `has_results: false` skills as "not yet evaluated" and suppress `verdict` for them (some checked-in reports are stale — eval re-runs requested from owners).
- `component` groups skills by product (from `components.d/`); 7 cuOpt skills currently have no component mapping (pre-consolidation leftovers).

## Status

- Schema v2 validated by the dashboard consumer (flat rows join onto adoption data on `catalog_dir`).
- Merge order: #325 must land first — it fixes the pre-existing metadata drift on main that currently fails the `check` job here before the new benchmarks validation step can run. Once CI is green end-to-end on this PR, it's ready.


mosheabr:
Note for reviewers: the failing `check` job is **pre-existing metadata drift on main**, not something introduced here. Commit 1f01acf (cuOpt orphaned-skill removal) landed without regenerating `metadata.json`/`skills.sh.json`, which still reference `cuopt-numerical-optimization-api-c` and `-cli`. This PR triggers the check because it touches the workflow file. Fix is a `workflow_dispatch` of the generate-skill-metadata workflow on main (or fold the regenerated files into a separate chore PR — happy to do either).

The new `benchmarks.json` drift gate added in this PR sits after that step, so it hasn't executed in CI yet; it passes locally against this branch (235 skills, byte-stable across runs).

.github/scripts/aggregate_benchmarks.py
```diff
@@ -0,0 +1,171 @@
...
+                    val = float(val)
+                entry[key] = val
+                break
+
+        if section == "agents used":
+            m = AGENT_LINE.match(line)
+            if m:
+                entry["agents"].append(m.group(1))
+
+        if section == "results" and line.startswith("|"):
```
mraheem-arch, line 66: The parser maps scores to agents positionally (zip(agents, cells[2:])), which trusts the results columns are [Dimension, count, agents-in-Agents-Used-order]. 
One table with a different column order or no count column and scores attach to the wrong agent. The results header already names the agent columns in backticks, so keying off the header row removes the assumption. 
Small change but it needs an agent_cols = {} init next to section = None too. 
Happy to send a patch if useful.  Just ping me

mosheabr: Good catch — fixed in 1e7b257. Agent columns are now parsed from the Results table header (backtick-stripped, Num excluded by name), with `agent_cols` init next to `section`, so reordered or partial columns attach to the right agent. Verified with a synthetic report using `| Dimension | codex | Num | claude-code |` ordering.

Same commit also switches the output to the flat rows layout from your Slack note: top-level `results[]` is one row per skill × dimension × agent, denormalized with `catalog_dir` + `component` for the adoption join; `skills[]` stays as a flat summary table with a `has_results` flag. schema_version bumped to 2.

mraheem-arch (approved):

mraheem-arch (approved):
re-approving post-rebase. pulled the regenerated benchmarks.json and diffed against the pre-rebase copy: only change is the 4 cuOpt orphans dropping out (235 -> 231 skills, 2080 result rows), header-keyed agent mapping intact (cudf values byte-identical), null-component count down to 8. rag-blueprint still has_results=false with the stale FAIL verdict, so we will gate verdict display on has_results dashboard-side as discussed.

sayalinvidia (approved):

PR #329, opened 2026-07-07
chore(plugins): point nvidia-skills bundle at consolidated cuOpt skill

jasonnvidia:
## Summary
- Swaps the curated cuOpt entry in `plugins.d/nvidia-skills.yml` `include_skills` from `skills/cuopt-numerical-optimization-api-python/` to the consolidated `skills/cuopt-numerical-optimization-api/`, following the same pattern as the user-rules → api-python swap in #305.
- Regenerates the plugin mirror (`build-plugins.py`): `plugins/nvidia-skills/skills/` now ships the consolidated skill instead of the old `-python` one.
- Structural change (curated skill set changed) auto-bumps the plugin minor version `1.1.2` → `1.2.0` via `version-plugins.py`, reflected in all three (Claude, Codex, Cursor) plugin manifests.

## Context
Upstream (NVIDIA/cuopt#1489) merged the three `cuopt-numerical-optimization-api-{c,cli,python}` skills into a single `cuopt-numerical-optimization-api`, now live in the catalog (synced via #322); the orphaned `-c`/`-cli` dirs were cleaned up in #323. The `-python` canonical dir couldn't be removed cleanly because the `nvidia-skills` bundle still curated it as the flagship cuOpt example, which would break `build-plugins.py` and the Validate Plugins drift check. With this include swap landed, the orphaned canonical `cuopt-numerical-optimization-api-python` dir can now be removed, completing the cuOpt consolidation end-to-end.

## Test plan
- [x] `build-plugins.sh --only nvidia-skills` regenerates the mirror cleanly (12 copies).
- [x] `version-plugins.sh --only nvidia-skills --base main --apply` reports the structural bump `1.1.2 → 1.2.0` and rewrites all three plugin manifests.
- [ ] CI `Validate Plugins` drift check is green on the regenerated tree.


mosheabr (approved):
Verified: plugins.d swap is exactly the api-python → consolidated-api change + 1.2.0 bump (minor, per the #305 structural-change precedent); regenerated mirror is byte-identical to the canonical skills/cuopt-numerical-optimization-api/ including signing artifacts; stale -python mirror dir fully removed; all three plugin manifests consistent at 1.2.0. CI green after approving the fork workflow runs. This closes the last leg of the cuOpt consolidation — the bundle had been shipping a deleted skill since the skills-dir cleanup.

PR #338, opened 2026-07-09
Attribute components for intentionally-unregistered skills (dashboard 'Unmapped' fix)

mosheabr:
Six skills exist without a components.d registration by design (contributor-facing `cuopt-developer`, five manually curated Physical AI Hub skills), so their `component` in benchmarks.json was null — rendering as **Unmapped** on the adoption dashboard's component view.

Fix: `catalog-exceptions.yml` entries may now declare a display `component`; the benchmarks aggregator uses it as a fallback (components.d stays authoritative where a registration exists). All 229 skills in the feed now carry a component (`cuopt-developer` → cuOpt; hub skills → Physical AI, per Alex Qi's single-product-entry convention). Regenerated output byte-stable.

Deliberately **not** done: registering these in components.d — that would put them in the sync's clone path (they're manually curated / contributor-facing) and surface cuopt-developer as a customer-facing product skill.

Reported by @mraheem-arch. cc @sayalinvidia

### DIFF
--- a/benchmarks.json
+++ b/benchmarks.json
@@ -186,7 +186,7 @@
         "codex"
       ],
       "catalog_dir": "cuopt-developer",
-      "component": null,
+      "component": "cuOpt",
       "has_results": true,
       "average_uplift_pct": 23.7
     },
@@ -2481,7 +2481,7 @@
       "verdict": "FAIL",
       "agents": [],
       "catalog_dir": "omniverse-cad-to-simready",
-      "component": null,
+      "component": "Physical AI",
       "has_results": false,
       "average_uplift_pct": null
     },
@@ -2496,7 +2496,7 @@
       "verdict": "FAIL",
       "agents": [],
       "catalog_dir": "omniverse-realtime-viewer",
-      "component": null,
+      "component": "Physical AI",
       "has_results": false,
       "average_uplift_pct": null
     },
@@ -2511,7 +2511,7 @@
       "verdict": "FAIL",
       "agents": [],
       "catalog_dir": "omniverse-usd-performance-tuning",
-      "component": null,
+      "component": "Physical AI",
       "has_results": false,
       "average_uplift_pct": null
     },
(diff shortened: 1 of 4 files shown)


benchmarks.json
jasonnvidia, line 1: Question on file placement: should `benchmarks.json` and `catalog-exceptions.yml` stay at the repo root, or would they fit better under a config folder? For example, marketplace metadata already lives under `.github/scripts/marketplace/`, so I’m wondering whether benchmark/dashboard metadata should follow a similar pattern or whether root-level discoverability is intentional here.

catalog-exceptions.yml
jasonnvidia, line 1: Does this overlap with `.github/scripts/manual-components.yml` which lists the manually staged Physical AI skills. Is that intentional? I guess should the benchmarks aggregator read from `manual-components.yml` or should we remove that file?

jasonnvidia (approved):
Should the benchmarks.json and catalog-exceptions.yml be a the top level or nested in a subfolder?

Also, since we already have .github/scripts/manual-components.yml with physical AI, does this overlap with the catalog-exceptions.yml?

mosheabr:
Both good questions — answered with a code change for the second:

**Placement:** root-level is intentional for `benchmarks.json` — it's a public consumer feed (the adoption dashboard and anyone else pulls it by raw URL), and it follows the `skills.sh.json` precedent of consumer-facing artifacts living at root. Internal-only build inputs stay under `.github/scripts/`. `catalog-exceptions.yml` needs to live outside any `components.d/*.yml` glob (that's the #334 lesson); root keeps it visible to the teams who need to edit it.

**Overlap: you were right, and it's now fixed in this PR.** `manual-components.yml` already declares the five Physical AI skills with their component name, so duplicating them in the exceptions file created a second source of truth. Latest commit: the aggregator and the orphan pruner now read `manual-components.yml` directly, the exceptions file shrinks to truly-uncataloged skills (just `cuopt-developer`), and its header documents the three-tier registry (components.d → manual-components → exceptions). The Physical AI team keeps their existing single edit surface. All local checks pass: zero null components, correct attribution, pruner clean.

(Also noting `manual-components.yml` says "TEMPORARY — remove after Computex 2026" — it's outlived that, but retiring it is the Physical AI team's repo-visibility call, tracked separately.)

PR #356, opened 2026-07-15
docs: add CLI-version requirement + install troubleshooting (VDR item 2)

mosheabr:
## What & why

The Verified Skills VDR flagged that skills install but do not appear in **Claude Code**. Root cause was an **upstream `skills` CLI bug, fixed in v1.5.16**: CLIs ≤ 1.5.15 skip creating the `.claude/skills` link, so files land in `.agents/skills/` but Claude Code never sees them. Codex reads `.agents/skills/` directly and was unaffected — which is why the VDR saw "works for Codex, not Claude Code."

No code/repo change is needed (the fix is upstream); the gap is **docs telling users to be on a current CLI**.

## Changes

**`README.md`**
- Quickstart callout: requires `skills` CLI **≥ 1.5.16** (use `npx skills@latest`), with a pointer to Troubleshooting.
- `/reload-skills` tip to load newly installed skills in-session.

**`docs/advanced-install.mdx`**
- New Requirements bullet for the ≥ 1.5.16 CLI floor.
- New **Troubleshooting** section: check CLI version first, reload tip, and a per-agent "verify your agent can see the skill" table (Claude Code vs Codex).

## Notes
- Addresses VDR action-plan **item 2** (docs). Item 1 (the symlink failure) is fixed upstream in CLI 1.5.16 — verified on macOS + x86_64 Linux.
- Docs-only; no skill content touched, so no re-sign needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,9 @@ npx skills add nvidia/skills
 
 The CLI runs through `npx` and prompts you to choose a skill and install destination. You do not need to clone this repo or copy skill folders by hand.
 
-The skill is available the next time your agent loads skills and encounters a relevant task. For example, ask your agent to "solve a linear programming problem with cuOpt" and the skill guides it through the cuOpt Python API.
+> **Requires a current `skills` CLI (v1.5.16 or newer).** Installing via `npx skills@latest add nvidia/skills` always uses the latest. On older CLIs (v1.5.15 and earlier), skills may install but not appear in Claude Code — see [Troubleshooting](docs/advanced-install.mdx#troubleshooting).
+
+The skill is available the next time your agent loads skills and encounters a relevant task. For example, ask your agent to "solve a linear programming problem with cuOpt" and the skill guides it through the cuOpt Python API. In Claude Code, run `/reload-skills` to load newly installed skills in your current session.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,7 +29,9 @@ npx skills add nvidia/skills
 
 The CLI runs through `npx` and prompts you to choose a skill and install destination. You do not need to clone this repo or copy skill folders by hand.
 
-The skill is available the next time your agent loads skills and encounters a relevant task. For example, ask your agent to "solve a linear programming problem with cuOpt" and the skill guides it through the cuOpt Python API.
+> **Requires a current `skills` CLI (v1.5.16 or newer).** Installing via `npx skills@latest add nvidia/skills` always uses the latest. On older CLIs (v1.5.15 and earlier), skills may install but not appear in Claude Code — see [Troubleshooting](docs/advanced-install.mdx#troubleshooting).
```
jasonnvidia, line 32: Maybe overkill to recommending to specify "@latest" here. Since we are using npx skills it will pickup the latest version when available (as long as skills isn't installed locally).

jasonnvidia (approved):

PR #383, opened 2026-07-27
Yashraj/fix signature guard status

yashrajp22:
The guard now:
- Recognizes only trusted, artifact-only signature commits.
- Checks the preceding content commit’s status.
- Waits up to 10 minutes if that status is still pending.
- Fails closed for unexpected actors/files.


mosheabr (approved):
Approving. Verified the signature-guard fix locally by running the guard script against mocked GitHub API responses across six scenarios: the previously-broken case (validated content commit + bot signature commit on top) now clears by inheriting the preceding content commit's status; two spoof attempts (non-bot actor, and a signature commit smuggling a SKILL.md) are correctly blocked; and the validated/unvalidated/no-watched-change baselines behave as expected. Logic is correct and spoof-resistant.

PR #398, opened 2026-08-02
fix(ci): make NVSkills status gate reusable

yashrajp22:
- Checks only the latest commit that changed watched paths.
- Lets unrelated later commits inherit that validation result.
- Waits while validation is pending.
- Fails a new watched-path commit with missing/failed status.
- Immediately passes an exact NVSkills bot signature commit containing only generated artifacts.
- Handles renamed files and paginated commit file lists.
- Cancels stale gate runs when the PR moves again.

### DIFF
--- a/.github/workflows/request-nvskills-ci.yml
+++ b/.github/workflows/request-nvskills-ci.yml
@@ -3,11 +3,14 @@ name: Request NVSkills CI
 on:
   issue_comment:
     types: [created]
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
   push:
 
 jobs:
   request:
     if: >
+      github.event_name == 'pull_request' ||
       (github.event_name == 'issue_comment' &&
         github.event.issue.pull_request &&
         startsWith(github.event.comment.body, '/nvskills-ci')) ||
@@ -17,6 +20,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: read
+      statuses: read
     uses: NVIDIA/skills/.github/workflows/team-request.yml@main
     secrets:
       NVSKILLS_CI_DISPATCH_TOKEN: ${{ secrets.NVSKILLS_CI_DISPATCH_TOKEN }}
--- a/.github/workflows/request-nvskills-ci.yml
+++ b/.github/workflows/request-nvskills-ci.yml
@@ -21,6 +21,6 @@ jobs:
       contents: read
       pull-requests: read
       statuses: read
-    uses: NVIDIA/skills/.github/workflows/team-request.yml@main
+    uses: ./.github/workflows/team-request.yml
     secrets:
       NVSKILLS_CI_DISPATCH_TOKEN: ${{ secrets.NVSKILLS_CI_DISPATCH_TOKEN }}
--- a/.github/workflows/request-nvskills-ci.yml
+++ b/.github/workflows/request-nvskills-ci.yml
@@ -21,6 +21,6 @@ jobs:
       contents: read
       pull-requests: read
       statuses: read
-    uses: ./.github/workflows/team-request.yml
+    uses: NVIDIA/skills/.github/workflows/team-request.yml@main
     secrets:
       NVSKILLS_CI_DISPATCH_TOKEN: ${{ secrets.NVSKILLS_CI_DISPATCH_TOKEN }}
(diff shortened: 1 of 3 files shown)


mosheabr:
Thanks for this. The paging fix, the rename fix, and the stale-run guard are all real improvements.

I pulled the script out of the workflow and ran it against fake data, comparing this branch against what's on `main` today with identical inputs. Two things I'd like to understand before approving.

## 1. A failed validation now comes out green

These PRs have two kinds of commits: content commits, where a person edits a skill, and signature commits, where the bot adds signature and report files afterward. The check is meant to confirm the content passed validation.

What I tested:

1. A commit edits `skills/foo/SKILL.md`.
2. Validation runs and reports failure on it.
3. The bot pushes its signature commit on top.

On `main`, the check skips past the bot's commit, looks at the content commit underneath, sees the failure, and fails — with a link to the failed run.

On this branch, the check sees the bot's commit on top and passes immediately. It never looks at step 2. Two API calls and it's done. The failing validation is still there in the PR; the check just reports green.

The same happens when validation never ran at all.

Nobody can fake this — I confirmed a human pushing an identical commit still fails, so it isn't a security hole. What I want to check is the reasoning: it looks like the assumption is that the bot only signs after a clean run, so its signature is proof enough. Is that right? What makes me unsure is that `skill-card-review-needed.md` is on the list of files a signature commit may contain — which reads like the bot does sign when something still needs a human look. If it does, those pass now too.

## 2. Nothing uses the reusable trigger yet

The new `workflow_call:` trigger has no inputs and nothing in the repo calls it, so it does nothing as merged. One thing to watch for later: when a workflow is called from another workflow, the check's name changes. If this check is ever set as required under its current name, switching to the called version would quietly stop it from being required. Is there a follow-up PR that uses this?

## Two smaller things, neither blocking

**It can tie up a build machine for four hours.** If validation gets stuck, it checks every 30 seconds until it gives up — 480 checks, exactly four hours, around 484 API calls. Those calls come out of a budget shared with every other PR running this same check. If that budget runs out, the job dies in a way that looks like the check failed rather than something going wrong behind the scenes. Half an hour would cover any real run.

**It got slower for ordinary PRs.** Today the check asks "does this PR touch skills at all?" once and stops. This branch fetches every commit one by one to find out — so a 12-commit PR that touches no skills goes from 1 API call to 14 to reach the same "nothing to check" answer. Keeping the old single check as a quick exit first would fix that.

## How I tested

I extracted the shell script from the `run:` block of both this branch and `main`, stubbed out `curl` and `sleep`, and drove both against the same fixtures in a bash 5 container. Summary of what came back:

| Scenario | `main` | this branch |
| --- | --- | --- |
| Watched change, validation green | pass | pass |
| Watched change, validation **failed**, bot signs on top | fail | **pass** |
| Watched change, validation never ran, bot signs on top | fail | **pass** |
| Same, but a human pushes the identical commit | fail | fail |
| Skill moved out of `skills/` (rename) | reports `docs/foo.md` | reports `skills/foo/SKILL.md` |
| 12-commit PR touching nothing watched | 1 API call | 14 API calls |
| Validation stuck pending | fails after 4 calls | 484 calls, 480 waits, 4 hours |

Happy to share the harness if it's useful for iterating.

mosheabr (approved):
Approving.

The core change is replacing a trust heuristic with actual evidence. The old `is_trusted_generated_signature_commit()` exempted a commit when the pusher, the title, and the file shape all looked right, which is three guesses standing in for proof. Requiring a real successful `NVSkills CI` status at or after the latest watched-path commit is strictly stronger.

Two fixes here that were not part of the stated scope and are worth calling out:

- `get_commit_json()` pages the commit file list. The previous code read `.files` from a single response and failed closed at 100 or more files, so any skill with more than 100 files could never be classified.
- The stale-head guard exits cleanly when the PR has advanced past the SHA the run started on, instead of publishing a verdict about an old tree.

Bounded polling with `timeout-minutes: 65` and curl retries are sensible. One consequence worth knowing: a PR can now wait up to an hour on this gate, and while signing is down that wait will always end in failure at the timeout.

One rollout note, not a blocker. This adds a `pull_request` trigger and `statuses: read` to the template `request-nvskills-ci.yml`, but team repos hold their own copies of that file. I checked `physical-ai-data-factory` and its copy predates this change, so those teams will not pick up the gate until they update. Worth a note to teams so the improvement actually reaches them.

Merging this first; #402 is stacked on top of it.

PR #408, opened 2026-08-03
fix(metadata): never drop a skill when AI enrichment fails

mosheabr:
## Summary

A skill classified `materially_changed` is sent back to the AI client to see whether its existing metadata values should be amended. At that point every required field already holds a valid value, so the amendment is optional — but an `EnrichmentError` returned `None`, dropping the skill from `metadata.json` and `skills.sh.json` entirely.

The generator does report this. It prints a `PARTIAL SUCCESS` block naming the skill and the reason. What it does not do is stop: it exits 0, writes the reduced file, and the regeneration PR is opened as normal. Nothing downstream marks the loss — the output is schema-valid and internally consistent, so `--check` passes and the round-trip validation is explicitly handed the missing path as `skipped_paths`. The result is a published skill dropped from the marketplace by a PR whose diff looks like ordinary churn.

## How it showed up

`cuopt-multi-objective-exploration` was legitimately `materially_changed` on 2026-08-03 after the 15:04 sync picked up upstream edits. Two generator runs three minutes apart, on the same commit, with the same classification (`materially_changed: 1`):

| Run | Trigger | Result |
| --- | --- | --- |
| [30842167516](https://github.com/NVIDIA/skills/actions/runs/30842167516) | `workflow_run` | `metadata.json: updated (323 skills)` — entry kept |
| [30842382996](https://github.com/NVIDIA/skills/actions/runs/30842382996) | `workflow_dispatch` | `metadata.json: updated (322 skills)` — entry dropped |

The trigger was a transient API response, not anything about the skill:

```
PARTIAL SUCCESS: 1 skill(s) could not be enriched and were excluded from output:
  - skills/cuopt-multi-objective-exploration: Inference API returned HTTP 403: 403 Forbidden
```

The identical call had succeeded three minutes earlier. `403` was not in the retryable set, so it was terminal on the first attempt. The run then opened a regeneration PR whose diff deleted the skill's 12-line entry from `metadata.json` and its name from `skills.sh.json`. It was caught by eye in review.

## Changes

1. **Amendment path keeps the existing entry.** On `EnrichmentError`, record a warning and carry the current metadata forward rather than returning `None`. Output stays byte-stable when the AI returns the same values, as before.
2. **A dropped skill fails the run.** If a discovered skill produces no entry — and it was either in the previous `metadata.json` or AI enrichment was available — the run errors and names it. Skills removed via `metadata-exclusions.yaml` are filtered out in `discover_skills` and never reach this point, so deliberate removals are unaffected. Since `Open or update validation issue` already fires on a non-zero exit, this also gives the case a tracking issue without any new wiring.
3. **`403` is now retried**, alongside `429` and `5xx`. The gateway fronting the inference API returns 403 for transient edge rejections as well as for genuine permission failures. Retrying costs roughly 7s on a real permission failure and saves an enrichment on a transient one.

Together: (3) makes the failure rarer, (1) makes it non-destructive when it still happens, (2) makes it loud and tracked if an entry genuinely cannot be built.

## Validation

Run against the real catalog with a stubbed inference API, comparing `main` against this branch:

| Scenario | `main` | this branch |
| --- | --- | --- |
| 403 then 200 | terminal after 1 call | recovered on call 2, 5 fields returned |
| persistent 403 | 1 call, message doesn't cite attempts | 4 calls, terminal, message cites attempts |
| amendment path, enrichment raises | entry dropped | entry kept, 5 metadata fields intact |
| forced drop, write mode | `updated (322 skills)`, exit 0 | `VALIDATION FAILED` naming the skill, exit 1 |

The fourth row reproduces run 30842382996 exactly.

Normal run, no induced failure — no behaviour change:

```
metadata.json: unchanged (323 skills)
skills.sh.json: unchanged (18 non-empty groups)
--check exit 0          (metadata)
--check exit 0          (benchmarks)
```

## Note

This does not explain *why* the gateway returned 403 on that call. It makes the failure rarer, non-destructive, and visible; if the 403s themselves are worth chasing, the two run IDs above are the starting point.

### DIFF
--- a/.github/scripts/marketplace/generate-skill-metadata.py
+++ b/.github/scripts/marketplace/generate-skill-metadata.py
@@ -1115,10 +1115,16 @@ def main(argv: list[str] | None = None) -> int:
             json.dumps({"stale_skills": skill_warnings}, indent=2) + "\n",
             encoding="utf-8",
         )
-    # A stale entry still has to fail the run. Outputs are written first so the
-    # catalog does not lose a skill, but the exit code stays non-zero in both
-    # --check and write mode so CI cannot read a partial success as a pass.
-    exit_code = 1 if skill_warnings else 0
+    # A skill that could not be built has to fail the run, in both --check and
+    # write mode, so CI cannot read a partial success as a pass. Outputs are
+    # still rendered first, so the catalog does not lose a skill either way.
+    #
+    # --no-ai is the exception. PR CI runs without an inference key on purpose,
+    # so a genuinely new skill always lands there with its fields unenriched;
+    # filling them is the regenerate job's work, not the PR's. Failing here
+    # would red-line every open PR from the moment a sync adds a skill until
+    # regeneration runs. The warnings are still printed either way.
+    exit_code = 1 if skill_warnings and not args.no_ai else 0
 
     metadata_text = dumps_canonical(metadata_obj)
     skills_sh_text = dumps_canonical(skills_sh_obj)
--- a/.github/scripts/marketplace/generate-skill-metadata.py
+++ b/.github/scripts/marketplace/generate-skill-metadata.py
@@ -1115,6 +1115,10 @@ def main(argv: list[str] | None = None) -> int:
             json.dumps({"stale_skills": skill_warnings}, indent=2) + "\n",
             encoding="utf-8",
         )
+    # A stale entry still has to fail the run. Outputs are written first so the
+    # catalog does not lose a skill, but the exit code stays non-zero in both
+    # --check and write mode so CI cannot read a partial success as a pass.
+    exit_code = 1 if skill_warnings else 0
 
     metadata_text = dumps_canonical(metadata_obj)
     skills_sh_text = dumps_canonical(skills_sh_obj)
@@ -1140,7 +1144,7 @@ def main(argv: list[str] | None = None) -> int:
             return 1
         _print_classification(cls)
         print("OK: generated outputs are byte-stable.")
-        return 0
+        return exit_code
 
     md_changed = write_if_changed(METADATA_PATH, metadata_text)
     sh_changed = write_if_changed(SKILLS_SH_PATH, skills_sh_text)
@@ -1155,7 +1159,7 @@ def main(argv: list[str] | None = None) -> int:
         f"({len(skills_sh_obj.get('groupings', []))} non-empty groups)"
     )
 
-    return 0
+    return exit_code
 
 
 def _print_classification(cls: Classification, stream=sys.stdout) -> None:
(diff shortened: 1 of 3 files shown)


mosheabr:
Thanks, the pipefail catch was the real root cause. All four points are fixed, plus the test you asked for.

1. Added `set -o pipefail` to the workflow. `tee` was hiding the generator's exit code, so a failed run looked like a pass. That is what let the dropped skill through CI.
2. When a skill cannot be rebuilt, it now keeps its old entry from metadata.json instead of vanishing. The old entry is checked against the schema first. Only skills with nothing valid to fall back on get left out.
3. If the optional amendment step fails on a skill that was already complete, we no longer say it was omitted. The skill ships unchanged with the values it had, and we log a note instead of a warning.
4. Notices and warnings now print before `--check` returns, so they show up in PR CI too.

On testing, I landed it rather than just running it once. `test_generate_skill_metadata.py` makes enrichment fail on purpose using a stub inference API. It does not touch the network and does not write to the repo. It runs in the existing check job, 14 assertions, all green on this run. The test builds its fixtures from whatever is in metadata.json at the time, so renaming a skill will not break it.

I also checked the test can actually fail: if I turn the recovery off, 3 assertions fail. If I send amendment failures back into skill_warnings, 2 fail.

Three bugs turned up while doing this, all fixed here:

- My first version of the recovery check validated one entry against the schema for the whole file, so it never matched and nothing was ever recovered.
- Moving the warning block above the `--check` return accidentally dropped its `return 1`, which would have brought back the same kind of bug this PR is fixing.
- The issue-body script still read `warnings.omitted_skills` after I renamed the key to `stale_skills`, so it would have shown zero affected skills every time. Fixed, and reworded: these skills are stale, not excluded.

One judgment call worth calling out. The generator runs in two modes: PR CI runs it without an inference key, so it cannot call the model, and the regenerate job runs it with the key, so it can.

Missing metadata now only fails the run if the generator could call the model and still could not fill the fields in. If it was never able to call the model, missing fields are expected, since filling them is the regenerate job's work, not the PR's. Failing there would turn every open PR red until regeneration ran, which is exactly what CI on this branch did with the three new hsb-ip skills. Warnings still print in both modes. There are tests for both cases.

jasonnvidia (approved):
I re-ran the test suite locally — 14/14 pass.  lgtm, approved!

**Claude's Non-blocking nits** feel free to ignore
Test gap: case 4 asserts recovery but not rc == 1. The behavior is correct today; one assertion would pin the “fail loudly” half so it can’t silently regress to exit 0 + open a PR with stale entries.
Noisy warnings: fill-mode failure appends to skill_warnings, then recovery appends again for the same path — two stale-skill lines for one skill.

PR #410, opened 2026-08-03
Rename People Attribute Search skill to Image Attribute Augmentation

sauravn-hub:
Renames the People Attribute Search skill to **Image Attribute Augmentation** in the catalog registration.

Only the manual `components.d` entry is changed:
- `components.d/physical-ai-data-factory.yml` — `path` + `catalog_dir` moved together to `physical-ai-image-attribute-augmentation`.

`skills/` and `metadata.json` are intentionally left untouched: the hourly sync copies in the renamed folder from the source repo and prunes the old one, then the regeneration bot updates `metadata.json` from what's on disk. Editing `metadata.json` here would point the index at a folder not yet present and trip the drift check.

Pairs with the source-repo rename in NVIDIA/physical-ai-data-factory.

### DIFF
--- a/components.d/physical-ai-data-factory.yml
+++ b/components.d/physical-ai-data-factory.yml
@@ -6,7 +6,7 @@ skills:
     catalog_dir: physical-ai-defect-image-generation
   - path: skills/physical-ai-video-data-augmentation/
     catalog_dir: physical-ai-video-data-augmentation
-  - path: skills/physical-ai-people-attribute-search/
-    catalog_dir: physical-ai-people-attribute-search
+  - path: skills/physical-ai-image-attribute-augmentation/
+    catalog_dir: physical-ai-image-attribute-augmentation
 links:
   discussions: false


mosheabr:
Thanks — the upstream side is in good shape. In `NVIDIA/physical-ai-data-factory` the directory is now `physical-ai-image-attribute-augmentation`, the old one is gone, and the new path carries `SKILL.md`, `skill.oms.sig`, `skill-card.md`, `evals/` and a `BENCHMARK.md` reading `Overall verdict: PASS`. The frontmatter `name` matches the directory, and the `components.d` edit moves `path` and `catalog_dir` together, which is exactly right.

Two things before this can merge.

**1. Please drop the `metadata.json` change and keep only the `components.d` edit.**

`metadata.json` is generated, and it's generated from the folders physically present in `skills/` in this repo — not from `components.d`. The catalog hasn't caught up with the upstream rename yet: `main` still contains `skills/physical-ai-people-attribute-search/`, because no sync has run since the directory was renamed. So the new entry points at a folder that isn't here yet, and the generator wants to remove it. That's the `check` failure:

```
DRIFT DETECTED
-      "path": "skills/physical-ai-image-attribute-augmentation",
-      "name": "physical-ai-image-attribute-augmentation",
```

The content of your entry is fine — it's what the generator will produce by itself. The problem is timing: merged now, `main` carries an index entry for a missing folder, which turns the drift check red on every PR to this repo until a sync and a regeneration land. The regeneration bot would also strip the entry in the meantime, since it builds the index from what's on disk.

With just the `components.d` change, the sequence takes care of itself:

1. This PR merges.
2. The next hourly sync reads the new path, copies in `physical-ai-image-attribute-augmentation`, and prunes the old directory.
3. The regeneration bot opens its usual PR with `metadata.json` updated.

**2. DCO is failing** — the commit has no `Signed-off-by` trailer:

```
git commit --amend -s
git push --force-with-lease
```

`authors`, `content-integrity` and `require-nvskills-ci` all pass already.

**One thing worth confirming:** after the sync prunes the old directory, `physical-ai-people-attribute-search` disappears from the published index rather than redirecting, so anyone who installed the skill under the old name will find it gone. If that's expected for this rename, no action needed — flagging it in case you want a transition period.

sauravn-hub:
Thanks @mosheabr , i have made the needful changes. Please check.

mosheabr (approved):
Approving.

Verified against the source repo `NVIDIA/physical-ai-data-factory`: the renamed path exists and carries SKILL.md, BENCHMARK.md, skill.oms.sig, skill-card.md and evals/. BENCHMARK.md is a PASS with real Tier 3 results. The skill was genuinely re-signed rather than just moved: all four artifacts are in one commit and the signature subject is the new name. SKILL.md frontmatter matches the directory, and the description keeps "people attribute search" as a trigger keyword so existing searches still find it.

On the failing check: it is `aggregate_benchmarks.py --check`, not the metadata generator. The rename leaves the old on-disk directory unregistered, so its 11 benchmark rows regenerate with `"component": null` instead of `"Physical AI"`. That is the entire diff, and it clears once the sync swaps the directories and the regeneration bot runs.

I checked the assumption this PR depends on rather than taking it on faith. Pruning is skipped entirely if more than PRUNE_CAP=5 directories would be removed. There are 7 unregistered directories on disk, but 5 are covered by manual-components.yml and 2 by catalog-exceptions.yml, so the real orphan count is 0. This merge takes it to 1, well under the cap, so the old directory will be pruned on the next sync. Same pattern as the cufolio to portfolio-optimization rename in 006749e.

Separately, and not a blocker here: every components.d rename will fail this check the same way, because benchmarks.json takes its component mapping from components.d but its rows from what is on disk, and the two are necessarily out of step mid-rename. Worth fixing so the next rename does not look broken.

mosheabr:
Correcting my approval above.

I said the skill had been genuinely re-signed rather than just moved, on the basis that all four artifacts were in one commit and the signature subject carried the new name. That is not sufficient evidence, and it turns out to be wrong. I checked the label on the signature, not the digests underneath it.

`skill.oms.sig` does not cover the current content. 20 of the 22 signed files match; `SKILL.md` and `skill-card.md` do not. The sync caught it and dropped the skill, and since the old directory had already been pruned in the same run, the net effect would have been to remove the skill from the catalog.

Reverting in #413 until the source repo is re-signed. The rename is fine and will go back in once `/nvskills-ci` has run there.

PR #20, opened 2026-04-16
Update Model-Optimizer skill count from 4 to 5

sayalinvidia:
## Summary

- Model-Optimizer repo added a 5th skill
- Updates the count in both README.md and ci.yml to unblock all open PRs

## Test plan

- [ ] ci.yml check passes on this PR

PR #429, opened 2026-08-07
fix: dispatch NVSkills CI for renames out of watched paths

yashrajp22:
## Problem

`require-nvskills-status.yml` (the required-status gate) classifies a PR as requiring an `NVSkills CI` status when **either** `filename` **or** `previous_filename` of any changed file is under a watched path (`skills/`, `team-skills/`, `rules/team-rules/`, `plugins/`). The dispatch step in `team-request.yml` only checked `filename`.

A PR that renames a file **out** of a watched path (e.g. `skills/foo/SKILL.md` → `docs/foo.md`, i.e. unpublishing a skill) therefore deadlocks:

- the gate demands an `NVSkills CI` status (it matches `previous_filename`) and fails after the missing-status grace period;
- `/nvskills-ci` skips with *"no changes under watched paths"* and never dispatches the pipeline that would produce the status.

The PR is permanently blocked.

## Fix

Align the dispatch step's watched-path filter with the gate's: match `previous_filename` as well (the jq `watched` predicate is copied verbatim from `require-nvskills-status.yml`). The pipeline validates the **resulting worktree**, so a rename that leaves no publishable skill resolves to a successful *not required* status — the same path deletions already take.

| Change | Dispatch | Outcome |
|---|---|---|
| `docs/foo.md` → `skills/foo/SKILL.md` | yes | scan |
| `skills/foo` → `skills/bar` | yes | scan |
| `skills/foo/SKILL.md` → `docs/foo.md` | yes (was: **never**) | not required |
| unrelated rename/modify | no | — |

## Verification

- jq filter exercised against 16+ edge-case payloads (rename out/in/within, deletion, missing/null `previous_filename`, `skillsX/` prefix confusion, unicode paths, empty pages) — all classify per the table above; `jq -e` exit codes drive the intended branches.
- Structural check: the gate demands a status **only if** its PR-level files scan matches, and that scan now uses the identical predicate — gate-demands ⇒ dispatch-fires, so the deadlock class is closed.
- Deadlock reproduced end-to-end and dispatch-after-fix demonstrated with the same one-hunk change on the copy consumed by NVIDIA/nvskills-ci-test: [demo PR #20](https://github.com/NVIDIA/nvskills-ci-test/pull/20) — before: dispatch step skipped + gate failed with `state: missing`; after: `should_dispatch=true` and the dispatch call was issued.
- Downstream: `run_validation_tier.sh` exits 0 (`No content directories changed — skipping`) when the resulting worktree maps to zero skill dirs, so the dispatched run completes successfully.

### DIFF
--- a/.github/workflows/team-request.yml
+++ b/.github/workflows/team-request.yml
@@ -41,11 +41,7 @@ jobs:
           ACTOR: ${{ github.actor }}
         run: |
           set -euo pipefail
-          role_name="$(curl -fsSL \
-            -H "Authorization: Bearer ${GH_TOKEN}" \
-            -H "Accept: application/vnd.github+json" \
-            "https://api.github.com/repos/${REPO}/collaborators/${ACTOR}/permission" \
-            | jq -r '.role_name // ""')"
+          role_name="$(gh api "repos/${REPO}/collaborators/${ACTOR}/permission" --jq '.role_name // ""')"
           case "${role_name}" in
             admin|maintain) ;;
             *) echo "Requester must have maintain or admin permission"; exit 1 ;;
@@ -156,7 +152,7 @@ jobs:
       - name: Dispatch NVSkills CI
         if: steps.context.outputs.should_dispatch == 'true'
         env:
-          DISPATCH_TOKEN: ${{ secrets.NVSKILLS_CI_DISPATCH_TOKEN }}
+          GH_TOKEN: ${{ secrets.NVSKILLS_CI_DISPATCH_TOKEN }}
           REPO: ${{ github.repository }}
           PR_NUMBER: ${{ steps.context.outputs.pr_number }}
           REQUEST_HEAD_SHA: ${{ steps.context.outputs.head_sha }}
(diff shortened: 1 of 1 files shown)


mosheabr (approved):
Approving. Verified the diagnosis against `main` rather than taking it on trust:

- `require-nvskills-status.yml` matches `(.filename | watched) or (.previous_filename? | watched)`
- `team-request.yml` matched `.filename` only

So the deadlock is exactly as described. A rename out of a watched path — unpublishing a skill — leaves the gate demanding a status that `/nvskills-ci` will never dispatch, with no error explaining why. Copying the `watched` function verbatim into the dispatch makes the two agree by construction instead of by coincidence, which is the right shape of fix for a bug caused by two copies of the same rule drifting apart.

The `gh` CLI refactor preserves the permission gate exactly — same `role_name` lookup, same `admin|maintain` allowlist, same non-zero exit otherwise — and the empty-token guard survives the rename to `GH_TOKEN`.

One improvement worth noting that isn't in the description: the old code passed the dispatch token as a curl `-H "Authorization: Bearer ..."` argument, so it appeared in the process arguments. Passing it through `GH_TOKEN` for `gh` to read removes that exposure. Same for the permission check.

PR #442, opened 2026-08-12
fix(ci): ignore trusted signature artifact commits

yashrajp22:
## Summary

- Treat tightly constrained NVSkills signing commits as generated artifacts when locating the latest content commit that requires validation.
- Continue requiring a real successful `NVSkills CI` status on or after that content commit; generated commits do not create a green result by themselves.
- Show repository-neutral fork guidance only for actual fork pull requests.

## Root cause

The required-status workflow treated the service account's generated `BENCHMARK.md`, `skill-card.md`, and `skill.oms.sig` commit as a new watched-path content change. Since that generated commit did not have its own `NVSkills CI` status, an already validated PR was blocked as `missing`. The failure text also hardcoded `NVIDIA/skills` and appeared on same-repository PRs.

This was reproduced on [physical-ai-skill-hub-dev PR #272](https://github.com/NVIDIA-dev/physical-ai-skill-hub-dev/pull/272) and [its failing workflow run](https://github.com/NVIDIA-dev/physical-ai-skill-hub-dev/actions/runs/31602421696/job/94173545357).

## Safety

A commit is skipped only when all of these hold:

- exact configured signature commit title
- GitHub author or committer matches the configured bot/service account
- every changed path is an allowed generated artifact path
- at least one `skill.oms.sig` is present
- no renamed files are present

A later human commit touching a watched path still becomes the validation floor and must have a successful status.

## Validation

- Parsed the workflow YAML and embedded Bash successfully.
- Replayed PR #272 against the live GitHub API: the gate passed using validated content commit `19c60a0876ab7d4e8ecd1579cfa20280ea5a580d` beneath signing commit `48004197bdd8c9c721c814024903ec8fa40c7c96`.
- Replayed the same PR with a missing status: the gate failed and did not show fork guidance.
- Replayed fork PR `NVIDIA/nvskills-ci-test#17`: generated signing commits were skipped correctly, and a missing-status run showed the generic parent-repository guidance.
- Confirmed human-authored lookalikes, extra content files, missing signatures, renamed files, and non-exact titles are not exempt.

### DIFF
--- a/.github/workflows/require-nvskills-status.yml
+++ b/.github/workflows/require-nvskills-status.yml
@@ -28,6 +28,9 @@ jobs:
           HEAD_REF: ${{ github.event.pull_request.head.ref || '' }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha || github.event.sha || github.sha }}
           STATUS_CONTEXT: ${{ vars.NVSKILLS_STATUS_CONTEXT || 'NVSkills CI' }}
+          SIGNATURE_COMMIT_TITLE: ${{ vars.NVSKILLS_SIGNATURE_COMMIT_TITLE || 'Attach NVSkills validation signatures' }}
+          SIGNATURE_PUSH_ACTOR: ${{ vars.NVSKILLS_SIGNATURE_PUSH_ACTOR || 'nv-skills-ci[bot]' }}
+          SIGNATURE_SERVICE_ACCOUNT_LOGIN: ${{ vars.NVSKILLS_FORK_SERVICE_ACCOUNT_LOGIN || vars.NVSKILLS_SIGNATURE_COMMIT_LOGIN || 'svc-nvskills-signing' }}
           STATUS_POLL_SECONDS: ${{ vars.NVSKILLS_STATUS_POLL_SECONDS || '30' }}
           MISSING_STATUS_GRACE_SECONDS: ${{ vars.NVSKILLS_MISSING_STATUS_GRACE_SECONDS || '120' }}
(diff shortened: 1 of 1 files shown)


keshprad (approved):

mosheabr (approved):
Approving.

Confirmed the exemption is currently absent from `main`, so the bug is real: the signing bot's generated commit is treated as a fresh watched-path change, has no `NVSkills CI` status of its own, and blocks an already-validated PR as `missing`.

Worth recording that this is the third move for this logic. #398 removed the exemption, #416 re-added it as part of fork support, #422 reverted #416 and took it with it. I reviewed #398 and called removing the exemption strictly stronger — that was wrong. It holds in this repo, where `request-nvskills-ci.yml` re-triggers on the bot's signature push, but not in team repos, which is exactly where #272 broke.

The safety conditions are tighter than what was there before: exact title, author or committer matching the configured account by resolved GitHub login rather than git metadata, every path an allowed generated artifact, at least one `skill.oms.sig`, and no renames. Keeping a later human commit on a watched path as the validation floor is the property that stops this becoming a general bypass.

The validation is the part that earns the approval. Replaying #272 against the live API both ways — passing on the correct content commit, and failing without fork guidance when the status is missing — plus confirming that human lookalikes, extra content files, missing signatures, renamed files and non-exact titles are all still caught. Testing the bypass attempts rather than just the happy path is what makes this reviewable.

Also picks up the hardcoded `NVIDIA/skills` in the failure text, which was showing fork guidance on same-repo PRs.

PR #453, opened 2026-08-13
chore: stop watching plugins/ and rules/team-rules/ for NVSkills CI

yashrajp22:
Removes `plugins/` and `rules/team-rules/` from the NVSkills CI watched-path classification, per policy decision. The watched set becomes `skills/` and `team-skills/`.

- `team-request.yml`: dispatch filter no longer matches either prefix (via `filename` or `previous_filename`); skip message updated.
- `require-nvskills-status.yml`: both the per-commit (`first_watched_path`) and PR-level watched filters updated; failure message updated.

Rationale: `plugins/` in this repo is a **generated tree** built from `skills/` + `plugins.d/` — gating it double-validates content already validated at its source (`validate-plugins.yml` still checks the generated tree for drift). `rules/team-rules/` is removed from gating by the same policy decision.

**Deliberately kept:** the generated-artifact regex for trusted signature commits still accepts `plugins/**` signature files, so signature commits touching plugin artifacts keep classifying as trusted.

Both watched filters remain byte-identical between the dispatch and gate sides, preserving the gate-demands ⇒ dispatch-fires invariant from #429.

Sibling change: NVIDIA/nvskills-ci#63. **Merge order matters: this PR must merge before NVIDIA/nvskills-ci#63** — team repos take their gate from this repo's `@main` but dispatch to nvskills-ci`@main`, whose validator will skip unwatched-only PRs without posting a status. Gates must stop demanding before the center stops supplying.

E2E validation (plugins case): NVIDIA/nvskills-ci-test#23 — no gate run on PR open, `/nvskills-ci` skips without dispatch, merge state CLEAN.

### DIFF
--- a/.github/workflows/require-nvskills-status.yml
+++ b/.github/workflows/require-nvskills-status.yml
@@ -113,8 +113,7 @@ jobs:
                 (. // "") |
                 startswith("skills/") or
                 startswith("team-skills/") or
-                startswith("rules/team-rules/") or
-                startswith("plugins/");
+                startswith("rules/team-rules/");
               [
                 .files[]? |
                 select((.filename | watched) or (.previous_filename? | watched)) |
@@ -197,8 +196,7 @@ jobs:
                 (. // "") |
                 startswith("skills/") or
                 startswith("team-skills/") or
-                startswith("rules/team-rules/") or
-                startswith("plugins/");
+                startswith("rules/team-rules/");
               any(.[]; (.filename | watched) or (.previous_filename? | watched))
             ' >/dev/null; then
               has_watched_change=true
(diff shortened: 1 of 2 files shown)


mosheabr (approved):
Approving. The rules/team-rules rationale answers it — this isn't just redundant like plugins/, it's an unresolvable block: a team storing rules in rules/ fails the status, runs /nvskills-ci as instructed, and fails again because rules aren't signable under the external profile.

Verified independently that no components.d entry declares a path outside skills/, so nothing pulls rules or plugins from onboarded repos.

Merging ahead of nvskills-ci#63 to preserve the ordering — gates stop demanding before the centre stops supplying.

PR #468, opened 2026-08-20
fix(benchmarks): capture v3 provenance and block silent field loss

mosheabr:
## Onboarding type

## Summary

SkillEvaluator's report format changed in July. It stopped emitting `NVSkills-Eval profile` and `Pass threshold`, both of which `aggregate_benchmarks.py` reads, so those columns in `benchmarks.json` have been emptying one skill at a time as each is re-evaluated:

| Eval month | Skills | `profile` null | `pass_threshold_pct` null |
|---|---:|---:|---:|
| 2026-05 | 41 | 0% | 36% |
| 2026-06 | 138 | 0% | 2% |
| 2026-07 | 133 | 57% | 8% |
| 2026-08 | 29 | **100%** | **100%** |

Nothing failed. Regeneration succeeded, the schema stayed valid, `--check` passed, and the data degraded silently for six weeks.

## Neither field is recovered here

`profile` is absent from v3 reports. The only `50%` remaining is a glossary sentence that is **byte-identical across skills with 4, 5 and 18 tasks** — template prose, not a measurement. Scraping it would stamp an unverified pass threshold onto every skill, in the file whose purpose is recording provenance. Both stay `null` until SkillEvaluator emits them as real fields; a `NO_FABRICATION` note in the source records why, so the next reader does not "fix" it.

## What is added

v3 carries stronger provenance the parser was discarding:

- `evaluator_version` — which build produced the score
- `dataset_digest` — sha256 of the exact task set, unique per skill
- `validation_status` — pipeline outcome for the run

This uses the multi-pattern idiom already in place for `tasks`, so v1/v2 reports are unaffected and leave the new fields `null`.

## The generic guard

`null_rate_regressions()` compares a regeneration against the committed `benchmarks.json` and **refuses to write** when a field lost values, counting only skills present in both files so new additions cannot false-fire.

```
  profile: 105 -> 125 nulls

This usually means an upstream BENCHMARK.md format change stopped
emitting a field the parser reads.
Either teach the parser the new format, or re-run with
--allow-null-regressions if the field is genuinely retired.
```

`--allow-null-regressions` lets a deliberate retirement land without a code edit under pressure. This is the same failure shape as the 2026-08-03 disappearance of `cuopt-multi-objective-exploration`.

## Verification

Against the real 343-skill catalog:

- **Does not fire today** — regeneration writes normally, exit 0
- **Does fire on the drift** — stripping the profile line from 20 real reports is refused, exit 1
- **Escape hatch works** — same state lands with the flag, exit 0
- **No regression** — every pre-existing field byte-identical across all 341 previously-present skills
- Flags the exact fields lost in #466 (`profile` 104→105, `pass_threshold_pct` 58→59)

## Tests

First test coverage for `.github/scripts/`. 16 cases across two files, using two real `BENCHMARK.md` fixtures, one per format. Run with:

```
cd .github/scripts/tests && python3 -m unittest discover
```

## Also

Adds the SPDX header `aggregate_benchmarks.py` was missing — it was the only `.py` in `.github/scripts/` without one. Called out separately since it is unrelated to the fix.

## Follow-up not in this PR

Restoring `profile` and `pass_threshold_pct` needs SkillEvaluator to emit them again, or a decision to retire them deliberately.

## All PRs


jasonnvidia (approved):
I've reviewed [PR #468](https://github.com/NVIDIA/skills/pull/468) end-to-end — read the parser, exercised the tests, ran the parser against all 343 real BENCHMARK.md files in the catalog, and simulated the guard against the actual main-vs-PR benchmarks.json transition.

lgtm

PR #477, opened 2026-08-22
fix(versioning): accept initial version for newly introduced plugins

rootkiller6788:
## What

`version-plugins.py` auto-bumps catalog plugin versions and validates builder-set bumps. When a `plugins.d/<name>.yml` is newly introduced (no yaml at the base ref), `analyze_plugin` sets `base_version = head_version` and `builder_changed_version = True`, so `decide()` runs `validate_builder_version(base, head)` with equal versions and reports:

```
! foo: version did not increase: base 1.0.0 -> head 1.0.0 (must be strictly greater)
```

There is no prior version for a brand-new plugin to increase from — the initial version is the builder's stamp, and it is already required to be a strict `MAJOR.MINOR.PATCH` by `SemVer.parse` in `analyze_plugin`. The current behavior makes every plugin onboarding fail (advisory in `validate-plugins.yml`, but fatal in `sync-skills.yml`'s `version-plugins.sh --apply --base HEAD` step).

## Change

- Add an `is_new` field to `PluginAnalysis`, set when the plugin yaml is absent at the base ref.
- In `decide()`, accept newly-introduced plugins (verdict `accept`) instead of applying the monotonic-increase check, which only makes sense when a base version exists.
- Existing-plugin behavior is unchanged: non-monotonic builder edits and oversized major skips still fail; structural/content changes still auto-bump y/z.
- Add `.github/scripts/tests/test_version_plugins.py` (10 tests) pinning both the new-plugin fix and the pre-existing `decide()`/`build_plan()`/`SemVer` behavior.

### DIFF
--- a/.github/scripts/tests/test_version_plugins.py
+++ b/.github/scripts/tests/test_version_plugins.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for version-plugins.py decision logic.
+
+Focuses on the `decide`/`build_plan` layer, which is pure and needs no git
+or worktree access. The module file carries a hyphen, so it is loaded via
+importlib rather than a plain `import`.
+"""
+
+import importlib.util
+import sys
+import unittest
+from pathlib import Path
+
+_SCRIPTS = Path(__file__).resolve().parent.parent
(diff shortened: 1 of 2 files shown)


mosheabr:
Thanks for tracking this down and for the thorough test coverage — the fix itself looks correct (confirmed `SemVer.parse` strictness is still enforced for new plugins, and the existing-plugin path is untouched).

Closing for now though: we only have one plugin in `plugins.d/` today (`nvidia-skills`), and plugin support more broadly isn't live yet, so this bug isn't reachable in practice. Once we're actually onboarding new plugins, we'll pick this back up — feel free to reopen or resubmit at that point. One thing to fix if/when you do: the commit is missing its `Signed-off-by:` DCO trailer.

PR #479, opened 2026-08-24
Update omniverse-usd-performance-tuning to 0.4.1

jensjebens:
Two fixes to the validation gate, both behavioral:

- **The validation aggregate no longer reports PASS when nothing was validated.** A blocked, timed-out, or user-declined validator disposition could previously satisfy the completion gate without any check actually running, so a report where every validator was unresolved could still summarize as an overall PASS. The aggregate now requires every planned validation unit to have actually executed before it can report PASS — a single unchecked item is enough to withhold it.
- **The validator subprocess now starts on Windows.** Its child process environment is built from an explicit variable allowlist rather than a full copy of the parent environment, and that allowlist was missing several variables Windows requires to start a Python subprocess. It failed to launch on that platform as a result. The missing variables are now included; the allowlist approach itself is unchanged.

Full detail in the skill's `CHANGELOG.md`.

`benchmarks.json` regenerated to match the updated `BENCHMARK.md`; verified with `generate-skill-metadata.py --check --no-ai` and `aggregate_benchmarks.py --check`.


mosheabr:
Thanks @jensjebens — this is merged as #481.

`/nvskills-ci` only runs on branches in the parent repo, so the signing gate here could never pass from a fork. Rather than ask you to redo the work, I moved your commits (`fd7de029`, `0652b2e8`) onto a branch in NVIDIA/skills as-is. Your authorship and sign-off are unchanged; only the branch location differed.

The signing run then re-ran the evaluation and regenerated BENCHMARK.md, skill-card.md and skill.oms.sig, which left the `benchmarks.json` in your commits stale against the new report. I regenerated it. The published figures now come from the same run that produced the live signature — evaluation date moved to 2026-08-24 and average uplift to 32.8, with Security at 100% for Claude Code.

Review notes, all clean: DCO signed, author affiliation good, content-integrity passes, BENCHMARK.md reports `Overall verdict: PASS` with real Tier 3 results across both agents, and all seven new `.py` files carry correct `Apache-2.0` SPDX headers placed after the shebang. Nicely done on the headers — that's the thing that usually needs a second pass.

One for next time: leave `benchmarks.json` out of the PR. It's regenerated from the BENCHMARK.md files after signing, so a hand-committed copy goes stale the moment the signing run touches the report.

Closing this one as superseded.

PR #485, opened 2026-08-25
fix(benchmarks): retire the profile field, accept v3 threshold nulls

mosheabr:
## Why

The `Generate Skill Metadata` job has failed on every sync since #482 landed TAO's 37 re-signed skills, opening issue #483 and emailing on each run:

```
  pass_threshold_pct:  62 -> 99 nulls
  profile:            109 -> 146 nulls
```

Both counts rose by exactly 37 — the TAO skills. This is the null-loss guard from #468 working correctly, not a break.

**Root cause, checked against the evaluator source rather than inferred.** SkillEvaluator 0.2.1 (released 2026-08-24, the same day TAO re-signed) "redesigned `BENCHMARK.md` as a decision-first publication card." `src/skillevaluator/reporting/benchmark.py` now emits only: Skill, Evaluation date, Evaluator version, Agents, Tasks, Dataset digest, Attempts per task, Environment, Tier 3 evidence, Tier 3 live evaluation, Overall verdict, Overall Tier 3 lift, Dimension bands. Neither dropped field has an emission path.

## What this changes

The two fields need different answers.

**`profile` — retired.** v1/v2 emitted `- NVSkills-Eval profile: external`. NVSkills-Eval is an internal name and does not belong in a published report, so v3 dropping it was correct. Keeping a permanently-null column named after an internal tool is worse than removing it.

**`pass_threshold_pct` — kept, null on v3.** It is a genuine provenance value that v3 turned into template prose. The same `50%` sentence appears verbatim on skills with 4, 5 and 18 tasks, so scraping it would stamp one number on every skill regardless of what it was actually evaluated against. The `NO_FABRICATION` note in the parser already made this call; this PR leaves it standing. Asking SkillEvaluator to emit it as a real field again is a separate conversation, not blocked by this.

## What's in the diff

- `aggregate_benchmarks.py` — remove the `profile` pattern, with a comment recording why it is retired rather than merely absent
- `test_aggregate_benchmarks.py` / `test_cli_guard.py` — both used `profile` as their sample droppable field; switched to live fields so the guard tests still exercise a real regression
- `benchmarks.json` — regenerated once with `--allow-null-regressions` to land both changes

## Verification

- 16/16 tests pass
- A strict `--check` passes against the regenerated file, so **the guard stays armed** for the next format change
- 343 skills, 0 carrying a `profile` key, 99 with a null `pass_threshold_pct` (the v3 cohort)

### DIFF
--- a/.github/scripts/aggregate_benchmarks.py
+++ b/.github/scripts/aggregate_benchmarks.py
@@ -28,9 +28,10 @@ from pathlib import Path
 SUMMARY_FIELDS = {
     "skill": [re.compile(r"^- Skill: `?([^`\n]+)`?\s*$")],
     "evaluation_date": [re.compile(r"^- Evaluation date: (.+)$")],
-    # v1/v2 only. v3 dropped this line along with the NVSkills-Eval name; it
-    # stays None there rather than being inferred. See NO_FABRICATION below.
-    "profile": [re.compile(r"^- NVSkills-Eval profile: `?([^`\n]+)`?\s*$")],
+    # RETIRED. v1/v2 emitted "- NVSkills-Eval profile: external". v3 dropped
+    # the line along with the NVSkills-Eval name, which is internal and does
+    # not belong in a published report. The field is retired rather than
+    # carried as a permanently-null column named after an internal tool.
     "environment": [re.compile(r"^- Environment: `?([^`\n]+)`?\s*$")],
     # v3 provenance. Absent from v1/v2, which leave these None.
     "evaluator_version": [re.compile(r"^- Evaluator version: `?([^`\n]+?)`?\s*$")],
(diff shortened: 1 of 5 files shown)


jasonnvidia (approved):
Approve. Verified locally:

- 16/16 tests pass
- `--check` passes on the regenerated file (guard stays armed)
- `benchmarks.json`: 0 entries carry `profile`, 100 nulls on `pass_threshold_pct` (v3 cohort)
- No downstream consumer of either field elsewhere in the repo

The two-fields-two-answers split is right: retiring `profile` (internal name in a public card) and keeping `pass_threshold_pct` as honest `None` on v3 rather than scraping the byte-identical glossary sentence. The `NO_FABRICATION` note earns its keep.

Nits, all non-blocking:
- Description says 99 nulls; regenerated file has 100 after the cuOpt sync commit landed on the branch
- The two guard test files pick different sentinel fields (`pass_threshold_pct` vs `environment`); either would work in both
- Two housekeeping commits (metadata + cuOpt regen) aren't mentioned in the PR body

PR #487, opened 2026-08-26
docs: document Tier 1 SkillSpector scan scope

### LINKED ISSUE #401: Checklist scan and the validation gate run SkillSpector over different trees, and authors cannot reproduce the gate's scope
The checklist and the validation gate both run SkillSpector, on the same skill, over different trees. The author gets the wider scope and is told to resolve everything in it, including findings the gate never sees.

The Scanning section says:

> Run SkillSpector against the complete skill directory.
>
> Resolve critical and high findings before release.

Run that way, SkillSpector skips only `.git`, `__pycache__`, `node_modules`, `.venv`, `venv`, `.tox`, and `.pytest_cache`. `evals/` is scanned like any other directory.

The gate runs the same scanner, but SkillEvaluator's Tier 1 security validator stages a filtered copy of the skill first, dropping everything in `SCAN_EXCLUDED_DIRS`: `evals`, `results`, `versions`, their dot variants, and the usual build caches, under the comment "Directories to skip at any depth during Tier 1 file-tree walks (scan artifact roots, version snapshots, build caches)." SkillSpector is then handed a tree in which those files never existed.

Excluding `evals/` from the gate looks deliberate and reasonable. The problem is that the checklist asks for a scan that cannot be made to match it. The filtering lives in SkillEvaluator, and SkillSpector exposes no exclude, ignore, or scope option and reads no configuration file, so an author who knows about the difference still has no way to reproduce the gate's scope. Nearly every skill in this repo has an `evals/`, so this is not an edge case in the layout.

## How much of the catalog this touches

| Skills in this repo                                       | Count |
| --------------------------------------------------------- | ----- |
| Total                                                     | 324   |
| With an `evals/` directory                                | 321   |
| Shipping files under `evals/` beyond the dataset manifest | 32    |
| Shipping packaging metadata under `evals/`                | 1     |

Most skills ship only the dataset manifest, which SkillSpector already exempts from static pattern analysis, so most authors never notice. The exposure grows with how much a skill puts under `evals/`.

## What it looks like when a skill does use fixtures

One skill ships eleven small buildable Python projects under `evals/files/`, staged as fixtures by its eval dataset. Scanning the complete skill directory, as the checklist instructs, returned 23 findings, 22 of them inside `evals/files/`, and a verdict of CRITICAL / DO NOT INSTALL.

Pinning every fixture dependency left 21 findings and the same verdict, because the advisory lookup keys on package name rather than the declared constraint (NVIDIA/SkillSpector#318). The only configuration that cleared it was deleting `pyproject.toml` from the eleven fixtures, which brought the scan to 2 findings and 42/100 MEDIUM.

So the checklist item "Resolve critical and high findings before release" was satisfied by degrading test fixtures until they no longer built. Nothing about the skill's own instructions, scripts, or declared dependencies changed.

The gate never depended on any of it. Restoring all eleven manifests and running `skillevaluator validate` returns 11 of 11 checks passed, exit 0, quality A at 94.8, with no high or critical findings and no reference to the fixture paths anywhere in the report. The fixtures were degraded to satisfy a scan whose result nothing consumes.

## The same mismatch persists after publication

The fixture case above is what an author meets while preparing a skill. A second case appears afterwards, on the other side of publication, and it is harder to argue with.

Every skill in this repo ships a `skill.oms.sig`, written by the publishing pipeline rather than by the author, and it only exists once a skill has been signed. So an author running the checklist before release never sees it. Anyone re-scanning a released skill does. Running the checklist scan on eight published skills sampled across the catalog produced the same result in all eight: exactly four critical or high findings, every one of them on the signature, and none anywhere else.

The gate does not see that file either, since `skill.oms.sig` sits in skillevaluator's `SCAN_EXCLUDED_FILES`.

The findings are not cosmetic. Two of the eight sampled skills land on `DO_NOT_INSTALL`, and removing only the signature file moves both back to `CAUTION`:

| Skill                       | As published          | Signature removed |
| --------------------------- | --------------------- | ----------------- |
| `amc-run-video-calibration` | 53/100 DO_NOT_INSTALL | 31/100 CAUTION    |
| `vss-search-archive`        | 51/100 DO_NOT_INSTALL | 29/100 CAUTION    |

So the checklist's own scan tells an author to resolve critical and high findings on published skills, where the findings come from a pipeline-generated signature the author cannot edit, and where they are sufficient on their own to produce a do-not-install verdict. The underlying rule behavior is filed as NVIDIA/SkillSpector#336. What makes it a checklist problem is that the gate already excludes the file and the checklist scan has no way to.

## The checklist already asks for something the tooling cannot do

The Before Scanning section says:

> Test fixtures, examples, and generated files are either intentionally included or excluded.

There is no way to act on that. SkillSpector has no exclude, ignore, or scope option, and reads no configuration file, so an author cannot mark fixtures as excluded even having decided to. That gap is filed as NVIDIA/SkillSpector#335. As written, the checklist asks for a decision one section before giving an instruction that overrides it.

## Why this matters left alone

Authors are pushed toward changes that lower quality. The cheapest way to clear fixture findings is to delete or thin the fixtures, which costs evaluation coverage and buys nothing, on files the validation gate had already decided were out of scope.

The same scanner returns two verdicts and neither report says why. An author comparing their checklist scan against the Tier 1 result sees two different answers for one skill, with nothing in either output indicating that the difference is scope rather than substance, and no way to run the narrower one to check.

It discourages shipping fixtures at all. Skills are asked to ship eval datasets, and a dataset is only as useful as the fixtures it exercises. Right now the checklist charges a skill for shipping them, and the 32 skills already putting content under `evals/` are the ones who would pay first.

The two halves belong to different teams, and neither can resolve it alone. Either the checklist names the scope the gate uses, or SkillSpector gains a way to scope a scan and the checklist stands as written. Both paths are reasonable, and picking one is the part that has not happened.

## Environment

- Checklist text from `docs/release-checklist.mdx` in this repo
- SkillSpector 2.5.1
- Catalog counts measured over `skills/*/evals/` in this repo

atirna:
## Onboarding type

- [x] Other (catalog documentation)

## All PRs

## Other context (for non-onboarding PRs)

SkillEvaluator stages Tier 1 before invoking SkillSpector, excluding evaluation and generated artifact paths. The release checklist previously told authors to resolve a broader standalone scan, which can report findings the release gate never sees.

This documents the Tier 1 scope in the checklist and updates its related trust-pipeline, evaluation, and signing references. Authors can now use SkillEvaluator to reproduce the findings that determine the release gate without weakening evaluation fixtures.

Fixes #401

Verification:

- Matched the documented scope against SkillEvaluator's current `SCAN_EXCLUDED_DIRS` (10 entries) and `SCAN_EXCLUDED_FILES` (3 entries).
- `git diff --check`


mosheabr (approved):
Thanks for this — verified the documented Tier 1 scope directly against skillevaluator's SCAN_EXCLUDED_DIRS (10 entries) and SCAN_EXCLUDED_FILES (3 entries) in source, and it matches exactly. Docs are accurate and clearly written. All CI checks (authors, dco, content-integrity) are green now that the fork-PR run got approved. Merging.

PR #492, opened 2026-08-28
feat(versions): publish per-skill content provenance for consumers

### LINKED ISSUE #491: Publish per-skill version provenance so consumers can detect updates
## Problem

Consumers who mirror skills from this catalog have no way to learn that a mirrored skill changed. Someone mirroring 30 skills today gets no signal when 4 of them are updated next month, short of walking all 343 directories and diffing content themselves.

Today the catalog publishes:

- **0 releases, 0 tags**
- `.github/scripts/marketplace/metadata.json` — inventory and taxonomy, but no version, digest or date
- `skills.sh.json` — groupings by name only
- `benchmarks.json` — evaluation provenance, not content provenance
- The README "Version" column — markdown only, and **per component**, not per skill

So the only machine-readable statement of "what changed" is the sync PR diff.

## Why this is small

The sync workflow already computes `upstream_sha`, `upstream_short_sha` and `upstream_date` per component on every run, writes them to a temp file, renders them into the README, and discards the machine-readable form.

Per-skill content digests also already exist: every skill carries a `skill.oms.sig` whose manifest lists per-file SHA-256 digests. Coverage is currently **343/343**, all manifests parse, and every manifest's subject name matches its directory name.

## Proposal

Add a `versions.json` at repo root, generated from committed catalog state:

```json
{
  "$schema": "...",
  "skills": [
    {
      "name": "aiq-deploy",
      "path": "skills/aiq-deploy",
      "content_digest": "sha256:...",
      "last_commit": "abc1234",
      "last_modified": "2026-08-19"
    }
  ]
}
```

`content_digest` is derived deterministically from the signed manifest, so it describes exactly what is published. A consumer stores the digests for the skills it mirrors and compares on refresh.

Then cut **releases** on a fixed cadence when content changes, with notes listing changed skills. That gives consumers a pinnable ref and `releases.atom`, which GitHub's watch → "Releases only" turns into an actual notification rather than a poll.

## Why a separate file, not `metadata.json`

`metadata.json` is taxonomy — product, category, subdomain, audience. It changes rarely and part of it is AI-enriched. Content provenance changes hourly and is purely mechanical. Keeping them separate means:

- taxonomy stops churning on every content sync
- a provenance bug cannot reach the marketplace index
- provenance does not inherit the AI-enrichment failure path
- the file can live at repo root, where an external consumer can find it

## Correctness requirement

The feed must describe what is **published**, not what was attempted. Skills are reverted by the sync when content changes without a signature refresh. If the feed reports those as updated, consumers chase updates that were rolled back. Deriving the digest from the committed `skill.oms.sig` rather than from the upstream sha the sync observed handles this by construction.

It also needs a guard against silent shrinkage: if the skill count drops without a corresponding removal in `components.d`, refuse to write. A generated file that stays schema-valid with one fewer entry is exactly how a skill silently disappeared from `metadata.json` on 2026-08-03.

## Open questions

- Release cadence — daily, weekly, or on every content-changing sync?
- Should `versions.json` also carry the source repo and source commit per skill, or is the content digest sufficient?

mosheabr:
Closes #491

## Onboarding type

## Problem

Consumers that mirror skills get no signal when a mirrored skill changes. The catalog publishes **0 releases and 0 tags**; `metadata.json` has no version or digest; `skills.sh.json` is groupings by name; the README's Version column is markdown and **per component**. Detecting that 4 of 30 mirrored skills moved means walking 343 directories and diffing them.

## What this adds

`versions.json` at repo root — one entry per skill:

```json
{
  "name": "accelerated-computing-cudf",
  "path": "skills/accelerated-computing-cudf",
  "content_digest": "sha256:4a42469b28c81689dea379388d555f574035795773eec55f542d15cad637fd19",
  "last_commit": "1f6f560df6257218e4a27f9fd5fc940078bfd140",
  "last_modified": "2026-05-29"
}
```

A consumer stores the digests for what it mirrored and compares on refresh.

## Design decisions worth reviewing

**The digest comes from `skill.oms.sig`, not the upstream sha.** When content changes without a signature refresh, the sync reverts the skill and the *old* content stays published. Keying on the signed manifest describes what is actually published, so a reverted update never surfaces as an update. Coverage is 343/343, all manifests parse, and every subject name matches its directory.

**Sorted file list, NUL-separated.** Sorting keeps the digest stable across a re-sign that shuffles manifest ordering. The NUL stops `{name: "ab", digest: "c"}` and `{name: "a", digest: "bc"}` colliding. Both are pinned by tests.

**Removals are gated.** If a skill disappears from a regeneration the generator refuses to write unless `--allow-removals` is passed. A generated file staying schema-valid with one fewer entry is exactly how `cuopt-multi-objective-exploration` silently vanished from `metadata.json` on 2026-08-03.

**Separate from `metadata.json`, deliberately.** That file is taxonomy — slow-moving and partly AI-enriched. This is mechanical and changes hourly. Separating them stops taxonomy churning on every content sync, keeps a provenance bug away from the marketplace index, and keeps AI enrichment out of the path for data that has nothing to do with it. It also lives at repo root where an external consumer can find it.

## Validation

- **22 unit tests**, all passing — digest determinism and order independence, separator collisions, change detection on add/remove/rename/edit, removal detection, schema closure, serialization stability.
- Generator runs clean against the live catalog: 343 skills, ~4s, **343 unique digests** (no collisions), dates spanning 2026-05-28 → 2026-08-27.
- `--check` is idempotent immediately after `--write`.
- Removal guard verified end-to-end: refuses with exit 1 on an injected phantom entry, passes with `--allow-removals`.
- Workflow parses as YAML; all 9 `run:` blocks pass `bash -n`.

## One change reviewers should look at closely

The `check` job's checkout gains `fetch-depth: 0`. The commit fields read `git log` per skill directory, which a shallow clone cannot answer. Without this the new `--check` step fails on every PR.

## Follow-ups, not in this PR

- **Releases/tags.** Issue #491 proposes cutting releases when content changes, which gives consumers a pinnable ref and `releases.atom` — the closest thing to a push notification. Cadence is an open question, so it is left out here.
- The repo's unit tests under `.github/scripts/tests/` were not wired into CI before this PR. This adds a runner for the new test file only; `test_aggregate_benchmarks.py` still has none.

- [x] Author confirmations above are checked
- [ ] `components.d/<slug>.yml` entry valid — n/a
- [ ] `SKILL.md` frontmatter spec-compliant — n/a, no skill content touched
- [x] No new license or third-party dependency introduced

## All PRs


.github/workflows/generate-skill-metadata.yml
siqinvidia: Should versions.json be included in this change-detection list? generate_versions.py can update it while the other generated files remain unchanged. In that case, changed would incorrectly remain false, and the workflow would skip creating the regeneration PR.

siqinvidia: versions.json needed here

siqinvidia: versions.json needed here

siqinvidia (commented):
Minor comments for adding versions.json to detection change list

siqinvidia (approved):
Minor freshness issue: versions.json was generated before the latest skills landed on main, so it currently contains 343 entries rather than 350. The next metadata regeneration should update it (in next hourly sync), but refreshing it in this PR would keep the new version index consistent from the moment it is introduced.

PR #501, opened 2026-08-31
Add PAIDF Augmentation and Auto-Labeling skills

sauravn-hub:
## Onboarding type

- [x] New product onboarding (new `components.d/<slug>.yml` file)
- [ ] Other (catalog change, README fix, infrastructure, etc.)

## For new product onboarding — author affirmations

- [x] **Skills cleared for open source release** by the owning team, with all required approvals in place
- [x] **License selected:** Apache 2.0
- [x] **No new license or new third-party component** introduced beyond what the source repositories already carry
- [x] **Source repositories are public and under an NVIDIA-owned GitHub org**
- [x] `skills/` paths are used for both new entries

## Summary

Registers the PAIDF v1.1.0 component skills for automatic catalog synchronization:

- `NVIDIA/paidf-augmentation` → `skills/paidf-augmentation/`
- `NVIDIA/paidf-auto-labeling` → `skills/paidf-auto-labeling/`

Both source directories contain a `SKILL.md` with the prefixed skill name and `physical-ai-data-factory` service metadata.

Related source releases:

- https://github.com/NVIDIA/paidf-augmentation/pull/5
- https://github.com/NVIDIA/paidf-auto-labeling/pull/3

## Validation

- Parsed all component YAML files and checked required fields
- Confirmed all `catalog_dir` values are unique
- Confirmed both public source paths and skill names exist
- Ran `git diff --check`
- [x] All commits signed off with DCO

PR #507, opened 2026-09-01
feat: onboard PAIDF Orchestration component

deegeee:
## Summary
- Registers `components.d/paidf-orchestration.yml` for `NVIDIA/paidf-orchestration`, following the flat layout described in `components.d/README.md`.
- Adds 4 skills: `write-dag`, `orchestration-setup`, `event-video-generation-workflow`, `image-attribute-augmentation-workflow`, each mapped to a `paidf-orchestration-`-prefixed `catalog_dir` for discoverability alongside the many other components in this catalog.

## Test plan
- [ ] Sync workflow picks up the new component file and produces `skills/paidf-orchestration-*` entries with no orphan-pruning conflicts.
- [ ] README's Available Skills table regenerates with the new PAIDF Orchestration entries.


mosheabr:
Thanks for this. The component file itself is well-formed, but there are two things to fix before it can merge.

**1. All four skills have a signature mismatch on `BENCHMARK.md` (blocking)**

I verified each skill's `skill.oms.sig` against the current content in `NVIDIA/paidf-orchestration@main`. Every file matches except `BENCHMARK.md`, in all four:

| Skill | Files matching signature | Mismatch |
|---|---|---|
| `write-dag` | 4 / 5 | `BENCHMARK.md` |
| `orchestration-setup` | 8 / 9 | `BENCHMARK.md` |
| `event-video-generation-workflow` | 13 / 14 | `BENCHMARK.md` |
| `image-attribute-augmentation-workflow` | 14 / 15 | `BENCHMARK.md` |

If this merges as-is, the sync will drop all four skills rather than publish them — and because they're new, there's no previously-signed version to fall back on, so they simply won't appear in the catalog. `paidf-augmentation` hit exactly this earlier today.

The catalog checks here can't catch it: `content-integrity` validates this repo's tree, not the source repo's, which is why everything looks green.

The cause is sequencing — `BENCHMARK.md` was regenerated after the signing run, so the signature covers the earlier report. Fix is to re-sign in the source repo so the signature covers the current content: open a PR touching the skill directories and comment `/nvskills-ci`. Make sure nothing regenerates `BENCHMARK.md` after that job completes.

**2. `dco` is failing**

Commit `e122c260` is missing its `Signed-off-by` trailer:

```
git rebase --signoff origin/main && git push --force-with-lease
```

**Everything else checks out:** source repo is public NVIDIA-org and Apache-2.0, all four declared paths exist with complete artifacts, all four `BENCHMARK.md` report `Overall verdict: PASS`, `catalog_dir` values are unique and product-prefixed, and only `components.d/` is touched.

One optional note: #501 just landed using `Physical AI Augmentation` / `Physical AI Auto-Labeling` as component names so they group as one product family in the README. You may want `Physical AI Orchestration` here for consistency — filename would become `physical-ai-orchestration.yml`.

mohgupta-ship-it:
LGTM

mohgupta-ship-it (approved):
lgtm

mohgupta-ship-it:
@mosheabr merging this PR. in case of issues, we can revert

PR #516, opened 2026-09-01
ci: verify source-repo skills on onboarding PRs

### LINKED ISSUE #508: Verify source-repo signatures on onboarding PRs before merge
## Problem

Onboarding PRs add `components.d/<slug>.yml` entries pointing at skill directories in a source repo. Nothing in this repo's CI verifies that those source skills carry a `skill.oms.sig` matching their current content.

The result is a failure that only surfaces *after* merge: the sync reverts the skill (for an existing one) or **drops it entirely** (for a new one, since there's no previously-signed version to fall back on). The onboarding PR itself shows all checks green, because `content-integrity` validates this repository's tree — not the source repository's.

## Evidence

Both of these happened on 2026-08-31:

- **`paidf-augmentation`** — onboarded via #501 with all checks green. The next sync dropped it: `1 signed file(s) do not match skill.oms.sig — new skill, no prior version`. Verifying by hand, 11 of 12 signed files matched; `BENCHMARK.md` did not.
- **`paidf-orchestration`** (#507) — the same defect in all four declared skills, caught only because the signature was checked manually during review:

  | Skill | Files matching | Mismatch |
  |---|---|---|
  | `write-dag` | 4 / 5 | `BENCHMARK.md` |
  | `orchestration-setup` | 8 / 9 | `BENCHMARK.md` |
  | `event-video-generation-workflow` | 13 / 14 | `BENCHMARK.md` |
  | `image-attribute-augmentation-workflow` | 14 / 15 | `BENCHMARK.md` |

The common cause is sequencing: `BENCHMARK.md` is regenerated after the signing run, so the signature covers an earlier report. It is an easy mistake to make and currently invisible until after merge.

## Proposal

Add a check that runs on PRs touching `components.d/**`. For each declared `path`:

1. Fetch the skill directory from the source repo at the component's `ref` (default `main`).
2. Parse `skill.oms.sig` and extract the signed resource digests.
3. Recompute SHA-256 for each signed file and compare.
4. Fail with the specific file names that do not match.

The same pass can cheaply assert the other onboarding hard requirements, which are also only enforced by reviewer discipline today:

- `SKILL.md`, `skill.oms.sig`, and a skill card are present
- `evals/evals.json` is present
- `BENCHMARK.md` exists and reports `Overall verdict: PASS` with real Tier 3 results

## Acceptance criteria

- An onboarding PR whose source skill has content not covered by its signature fails CI, naming the offending files.
- A PR where every declared skill is correctly signed passes.
- The check reads the source repo read-only and needs no signing credentials.
- Failure output tells the author what to do: re-run signing at source so the signature covers current content.

## Why it is worth doing

Post-merge drops are expensive to diagnose. The catalog looks correct, the PR looks correct, and the skill silently never appears. Moving this left turns a confusing absence into an actionable CI failure at the point where someone can fix it.

mosheabr:
Closes #508.

## The gap

An onboarding PR adds `components.d/<slug>.yml` pointing at skill directories in a product repo. Nothing on the PR looks at them. `verify-content-integrity` checks *this* repository's tree, which is sound — the content being onboarded lives somewhere else and goes unexamined. The PR goes green on evidence unrelated to what is being merged.

The cost lands after the merge. The sync recomputes each signed file's digest and refuses the skill on any mismatch: an existing skill reverts to its last good version, and a **new** skill is dropped outright, having no earlier version to fall back to. The PR is closed by then, so the failure surfaces nowhere. The catalog looks correct and the skill is simply absent.

Both happened on 2026-08-31. `paidf-augmentation` (#501) merged green and was dropped by the next sync. `paidf-orchestration` (#507) carried the same defect in all four declared skills and was caught only because the signatures were verified by hand during review. In every case the file was `BENCHMARK.md`, regenerated after the signing run so the signature covered an earlier copy.

## What this adds

A `Verify Source Onboarding` check on PRs touching `components.d/**`. For each declared path it fetches the source at the component's `ref` and verifies:

- every file listed in `skill.oms.sig` is present and still hashes to its signed digest
- `SKILL.md`, `skill.oms.sig`, a skill card and `evals/evals.json` exist
- `BENCHMARK.md` reports an overall verdict of PASS backed by real measurements

**Blocking scope is limited to paths the PR adds.** Around 30 catalog skills already carry drift (#216, #357); failing a PR for drift its author neither caused nor can fix would make the gate an obstacle rather than a signal. Pre-existing paths are still checked and reported, never fatal.

Read-only, anonymous clone, no signing credentials — so it runs on fork PRs.

## Reuse

The content check calls `verify_content_integrity.verify_skill` rather than reimplementing digest comparison; it takes a directory and does not care whose repository that directory came from. The benchmark check calls `aggregate_benchmarks.parse_benchmark`, which already handles the v1/v2/v3 report layouts — matching the verdict string by hand reads v2's methodology bullet ("Overall verdict: PASS only when every configured dimension...") as a passing verdict on every v2 report.

Stdlib only, matching the other scripts in `.github/scripts/`.

## Verification

Replaying `paidf-orchestration` at `5e4fd6b` — its state while #507 was under review — flags `BENCHMARK.md` on all four skills and nothing else, reproducing the table in #508:

```
FAIL  event-video-generation-workflow         BENCHMARK.md: HASH MISMATCH
FAIL  image-attribute-augmentation-workflow   BENCHMARK.md: HASH MISMATCH
FAIL  orchestration-setup                     BENCHMARK.md: HASH MISMATCH
FAIL  write-dag                               BENCHMARK.md: HASH MISMATCH
```

Also exercised end-to-end against live repositories: a new component with a clean source passes; a declared path absent at the ref fails; an unresolvable `ref` fails; a new path pointing at a source with real drift (`NVIDIA/NeMo-Retriever`) fails with 8 named files; and an edit to an existing component whose source is drifted warns and exits 0 rather than blocking.

21 unit tests cover the per-skill checks and the added-vs-pre-existing classification, against fixture directories — no network.

## Known limits

- The source is checked at the moment CI runs. A team that re-signs, goes green, then pushes a commit touching a signed file can still land the defect; the sync remains the backstop. This moves the common case left, it does not make the class impossible.
- On fork PRs the check waits for a maintainer to click "Approve and run" under the current Actions policy, so feedback is on first reviewer touch rather than instant.

## Test plan

- [ ] Check runs and passes on a PR touching `components.d/**` with a correctly-signed source
- [ ] Check is skipped on PRs that touch no component files

### DIFF
--- a/.github/scripts/verify_source_onboarding.py
+++ b/.github/scripts/verify_source_onboarding.py
@@ -1,0 +1,387 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: CC-BY-4.0 AND Apache-2.0
+"""Verify source-repo skills before an onboarding PR is merged.
+
+An onboarding PR adds ``components.d/<slug>.yml`` pointing at skill
+directories in a product repo. Nothing on the PR looks at those directories:
+``verify_content_integrity.py`` checks *this* repository's tree, which is
+sound, while the content being onboarded lives somewhere else entirely. The
+PR therefore goes green on evidence that has nothing to do with what is
+being merged.
+
+The cost lands after the merge. The sync fetches the source directory,
+recomputes each signed file's digest, and refuses the skill on any mismatch:
+an existing skill reverts to its last good version, and a *new* skill is
+dropped outright, because there is no earlier version to fall back to. The PR
+is closed by then, so the failure has nowhere to surface. The catalog looks
+correct and the skill is simply absent.
+
+That happened twice on 2026-08-31. ``paidf-augmentation`` (#501) merged green
+and was dropped by the next sync; ``paidf-orchestration`` (#507) carried the
+same defect in all four declared skills and was caught only because someone
+verified the signatures by hand during review. In every case the offending
+file was ``BENCHMARK.md``, regenerated after the signing run so the signature
+covered an earlier copy — an easy sequencing mistake that is invisible until
+it is expensive.
+
+This script follows the pointer. For each skill directory a PR declares it
+fetches the source at the component's ``ref`` and checks that:
+
+  * every file listed in ``skill.oms.sig`` is present and still hashes to its
+    signed digest;
+  * ``SKILL.md``, ``skill.oms.sig``, a skill card and ``evals/evals.json``
+    exist;
+  * ``BENCHMARK.md`` reports an overall verdict of PASS backed by real
+    measurements.
+
+Blocking scope: only paths the PR *adds*. Around 30 catalog skills already
+carry signature drift (tracked in #216 / #357); failing a PR for drift its
+author neither caused nor can fix would make the gate an obstacle rather than
+a signal. Pre-existing paths are still checked, and reported, but never fail
+the build.
+
+Reads the source repository read-only over an anonymous clone. No signing
+credentials, and nothing here writes to the source repo.
+
+Exit code 0 = no blocking problems; 1 = a newly-added skill is not fit to
+merge.
+"""
+from __future__ import annotations
+
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+import aggregate_benchmarks as ab  # noqa: E402
+import verify_content_integrity as vci  # noqa: E402
+
+COMPONENTS_DIR = Path("components.d")
+SIG_NAME = vci.SIG_NAME
+
+# Required alongside the signature. A skill card is required too, but its
+# filename is not settled: all three spellings below are in active use in the
+# catalog today (skill-card.md 343, SKILLCARD.yaml 47, card.yaml 18), so any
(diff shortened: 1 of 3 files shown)


.github/scripts/verify_source_onboarding.py
```diff
@@ -0,0 +1,387 @@
...
+
+COMPONENTS_DIR = Path("components.d")
+SIG_NAME = vci.SIG_NAME
+
+# Required alongside the signature. A skill card is required too, but its
+# filename is not settled: all three spellings below are in active use in the
+# catalog today (skill-card.md 343, SKILLCARD.yaml 47, card.yaml 18), so any
+# one of them satisfies the requirement.
+REQUIRED_FILES = ("SKILL.md", SIG_NAME, "BENCHMARK.md", "evals/evals.json")
+CARD_NAMES = ("skill-card.md", "SKILLCARD.yaml", "card.yaml")
```
siqinvidia: Should this require exactly skill-card.md so it matches the hourly sync? https://github.com/NVIDIA/skills/blob/1518780c5cf9cd1f3779cb5595aac13aadcb8485/.github/workflows/sync-skills.yml#L359

siqinvidia (commented):
Minor comment on skill-card.md to match the hourly sync checks

siqinvidia (approved):
Approved. The new gate closes the source-onboarding validation gap by checking signatures, required artifacts, and evidence-backed benchmark results before merge.

PR #518, opened 2026-09-01
fix: remove physical-ai-image-attribute-augmentation skill

deegeee:
Removes the `physical-ai-image-attribute-augmentation` entry from `components.d/physical-ai-data-factory.yml` and its `skills/physical-ai-image-attribute-augmentation/` catalog copy.


deegeee:
Superseded by #519, which folds this change into a combined PR.

PR #519, opened 2026-09-01
fix: update paidf-orchestration source paths and catalog naming

deegeee:
Supersedes #518.

- Updates `components.d/physical-ai-orchestration.yml` `path:` fields to match the renamed source folders in `sdg-workflow`'s `github-mirror` branch (see sdg-workflow MR !241): `skills/paidf-orchestration-write-dag`, `skills/paidf-orchestration-setup`, `skills/physical-ai-event-video-generation`, `skills/physical-ai-image-attribute-augmentation`.
- Renames the last two `catalog_dir` entries to `physical-ai-*` to match.
- Removes the now-freed-up `physical-ai-image-attribute-augmentation` entry from `components.d/physical-ai-data-factory.yml` in the same PR, so there's no window where the catalog_dir is duplicated.


components.d/physical-ai-orchestration.yml
```diff
@@ -2,13 +2,13 @@ name: Physical AI Orchestration
 repo: NVIDIA/paidf-orchestration
 description: Build, run and monitor physical AI data factory pipelines for image and video generation and transformation.
 skills:
-  - path: skills/write-dag/
+  - path: skills/paidf-orchestration-write-dag/
```
siqinvidia, line 5: The four new paths do not yet exist in NVIDIA/paidf-orchestration@main

skills/physical-ai-image-attribute-augmentation/assets/configs/osmo/augmentation.yaml
siqinvidia, line 1: Please remove the direct deletion of skills/physical-ai-image-attribute-augmentation/ and let the sync workflow replace or prune the catalog content. The current deletion leaves benchmarks.json stale and causes the benchmark validation to fail

siqinvidia (changes requested):
Thanks for cleaning up these mappings. I found two issues before merge:
1. Please land the renamed source folders from sdg-workflow MR !241 in the public NVIDIA/paidf-orchestration repo first.
2. Please remove the direct deletion of skills/physical-ai-image-attribute-augmentation/. Since the same catalog_dir remains registered under Orchestration, the hourly sync will replace its contents with rsync --delete once the new source path is available, while separately pruning the retired paidf-orchestration-image-attribute-augmentation-workflow directory.

PR #540, opened 2026-09-08
Remove remaining retired i4h catheter skills

wyli:
## Onboarding type

- [ ] New product onboarding (new `components.d/<slug>.yml` file)
- [x] Other (catalog change, README fix, infrastructure, etc.)

## For new product onboarding — author affirmations

By submitting this PR, I confirm on behalf of my team:

- [x] **Skills cleared for open source release** by the owning team, with all required approvals in place
- [x] **License selected:** Apache 2.0 / CC-BY 4.0 / Dual (Apache 2.0 + CC-BY 4.0). Specify: Apache 2.0
- [x] **No new license or new third-party component** introduced beyond what the source repo already carries
- [x] **Source repo is public and under an NVIDIA-owned GitHub org**
- [x] `.agents/skills/` or `skills/` path used for new entries (or existing path retained for legacy entries per `components.d/<slug>.yml`)

## Reviewer checklist (OSS Skills PIC)

- [ ] Author confirmations above are checked
- [ ] `components.d/<slug>.yml` entry valid (required fields, unique `catalog_dir`, path exists in source repo, filename slug matches name)
- [ ] `SKILL.md` frontmatter spec-compliant (at least one sampled)
- [ ] No new license or third-party dependency introduced

## All PRs

- [x] All commits signed off with DCO (`git commit -s`).
      If you forgot, run `git rebase --signoff origin/main && git push --force-with-lease` to retroactively sign all commits in your branch.

### Summary

- Remove the final two catheter-navigation mappings whose source skill directories have been retired.
- Align the component registry with all 14 skills currently present under `isaac-for-healthcare/i4h-workflows/skills` on `main`.

AI-assisted: Created with Codex/GPT at the user's request.

PR #542, opened 2026-09-08
fix(versions): let a deregistered skill leave versions.json

mosheabr:
## Onboarding type

- [x] Other (CI / generator fix)

## All PRs

### The problem

`generate_versions.py` refuses to write when any skill present in the checked-in `versions.json` is missing from a regeneration. That guard is correct for the case it was written for — `cuopt-multi-objective-exploration` disappearing on 2026-08-03 while still registered — but it also fires on the routine case of a team retiring a skill.

It fired today. Isaac for Healthcare deregistered five catheter skills in c3168ca, the sync pruned them in #538, and the regeneration refused:

```
Refusing to write versions.json: 5 skill(s) disappeared from the output.
  - i4h-catheter-navigation
  ...
If these were deregistered in components.d, re-run with --allow-removals.
```

The workflow never passes `--allow-removals` and exposes no input to set it, so there is no way to record that a removal was intended.

Two knock-on effects worth naming:

1. **It strands the other three files.** This is the last generation step, so a non-zero exit skips "Create regeneration PR" — `metadata.json`, `skills.sh.json` and `benchmarks.json` were all generated correctly and then discarded.
2. **It cannot self-heal.** The comparison is against a checked-in file that only a successful regeneration can update, so every subsequent hourly run fails identically and re-comments on the tracking issue (#541).

### The change

Registration separates the two cases exactly:

- a skill absent from `components.d` was retired on purpose — allow it, and log which ones
- a skill still registered that disappears anyway — still blocks

This is the same distinction #522 drew in `aggregate_benchmarks.py` for the same class of false alarm, and it reuses `registered_catalog_dirs()` rather than restating the rule.

An empty registry is treated as a parse failure rather than as "nothing is registered", so a bad parse cannot silently switch the guard off. `--allow-removals` is unchanged as a manual override.

The decision is factored into `blocking_removals(gone, registered)` so it can be tested directly, matching how the rest of this module is covered.

### Verification

Against `main` at 0b26974:

| Case | Before | After |
|---|---|---|
| 5 deregistered skills removed (today's failure) | refuses, exit 1 | writes 346 skills, names the five, exit 0 |
| Still-registered skill vanishes (2026-08-03 shape) | refuses | **still refuses**, exit 1 |
| `components.d` parse returns empty | n/a | all removals block — guard not disabled |

Confirmed the corrected output keeps `-smoke`/`-e2e` (still registered, handled separately in #540) and includes the newly synced `i4h-workflow-train-rl`.

27 tests pass, including five new ones pinning both directions.

`versions.json` itself is deliberately not touched here — the bot regenerates it with AI enrichment, which is required for skill classification, once this lands.

## Test plan

- [ ] Post-merge regeneration completes and refreshes #539 with correct content
- [ ] A future deregistration (e.g. #540's `-smoke`/`-e2e`) no longer wedges the hourly run

### DIFF
--- a/.github/scripts/generate_versions.py
+++ b/.github/scripts/generate_versions.py
@@ -37,6 +37,8 @@ from pathlib import Path
 
 import jsonschema
 
+import aggregate_benchmarks as ab
+
 REPO_ROOT = Path(__file__).resolve().parents[2]
 SKILLS_DIR = REPO_ROOT / "skills"
 OUTPUT = REPO_ROOT / "versions.json"
@@ -112,6 +114,24 @@ def removed_skills(new: dict, old: dict) -> list[str]:
                   - {s["name"] for s in new["skills"]})
 
 
+def blocking_removals(gone: list[str], registered: set[str]) -> list[str]:
+    """Removals that must stop the write, given the registered skill set.
+
+    A skill whose components.d entry was dropped is *expected* to leave the
+    output: the sync prunes its directory and versions.json follows. The case
+    this guard exists for is the opposite one -- a skill still registered that
+    vanished anyway, which is how cuopt-multi-objective-exploration
+    disappeared on 2026-08-03. Registration is what separates them.
+
+    An empty ``registered`` means the parse failed rather than that nothing is
+    registered, so nothing is exempt; otherwise a bad parse would silently
+    switch the guard off.
+    """
+    if not registered:
+        return list(gone)
+    return [name for name in gone if name in registered]
+
+
 def serialize(doc: dict) -> str:
     return json.dumps(doc, indent=2, ensure_ascii=False) + "\n"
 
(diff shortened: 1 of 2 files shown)


jasonnvidia (approved):
lgtm, from what I can tell this is the new flow:
```mermaid
flowchart LR
    subgraph Intentional["Intentional (allowed automatically)"]
        direction TB
        I1[Team drops entry<br/>in components.d] --> I2[Sync prunes<br/>skills/dir]
        I2 --> I3[Skill absent from output<br/>AND not registered]
        I3 --> I4["note: detected N<br/>deregistered skill(s)"]
        I4 --> I5[Write proceeds]
    end
    subgraph Silent["Silent loss (blocked)"]
        direction TB
        S1[components.d entry<br/>still present] --> S2[Skill vanishes<br/>from output anyway]
        S2 --> S3[Absent from output<br/>BUT still registered]
        S3 --> S4[Refuse to write, exit 1]
        S4 --> S5["Fix registration,<br/>or --allow-removals override"]
    end
```

PR #570, opened 2026-09-15
fix(metadata): exempt validation_status from the null-regression guard

### LINKED ISSUE #569: [skill-metadata] generator validation failed
Validation failed in workflow run: https://github.com/NVIDIA/skills/actions/runs/35006230897

Trigger: `workflow_run`

cc @jim-NVIDIA @jasonNVIDIA

<details><summary>Full generator log</summary>

```
Skill change classification:
  added: 10
    - skills/bionemo-kermt-add-cmim-pretrain
    - skills/bionemo-kermt-continue-pretrain
    - skills/bionemo-kermt-embed
    - skills/bionemo-kermt-finetune
    - skills/bionemo-kermt-infer
    - skills/bionemo-kermt-monitor
    - skills/bionemo-kermt-pretrain-scratch
    - skills/bionemo-kermt-setup
    - skills/foundationpose-pipeline
    - skills/foundationpose-setup
  removed: 0
  renamed: 0
  materially_changed: 1
    - skills/nemotron-speech
  excluded: 0
  unchanged: 354
metadata.json: updated (365 skills)
skills.sh.json: updated (17 non-empty groups)
note: pass_threshold_pct 171 -> 173 nulls — expected while cards migrate; see MIGRATING_FIELDS.
Skills carried these before and would not after:
  validation_status: 248 -> 259 nulls

This usually means an upstream BENCHMARK.md format change stopped emitting a field the parser reads.
Either teach the parser the new format, or re-run with --allow-null-regressions if the field is genuinely retired.

```

</details>

mosheabr:
## Problem

`Generate Skill Metadata` has failed on every hourly run on `main` since 2026-09-15 18:14Z (issue #569):

```
Skills carried these before and would not after:
  validation_status: 248 -> 259 nulls
```

SkillEvaluator 1.5.x stopped emitting the `- Validation status: \`passed\`` line that `aggregate_benchmarks.py` parses. The #560 sync landed 10 new skills (8 BioNeMo KERMT, 2 FoundationPose) plus a re-signed `nemotron-speech`, all on 1.5.x, so 11 more skills lost the field and the guard blocked the write. It will block on every future re-sign until resolved.

## Fix

Add `validation_status` to `MIGRATING_FIELDS`, the mechanism already used for `pass_threshold_pct`, which has the identical story (v1/v2 emitted it, v3 dropped it, each migrated skill adds one null).

The drift is still reported — it just becomes a note rather than a blocking failure. **Every other field stays guarded**, so a parser that genuinely breaks is still caught.

## Why not derive it from another field

Considered and rejected, deliberately:

- **The line carries exactly one value.** All 96 cards that still emit it say `passed`; none has ever said anything else. It distinguishes nothing, so there is no signal to preserve.
- **Those 96 cards are precisely the 96 still carrying the internal CI image path** — i.e. the set last signed before 1.5.4. Verified: the two sets are identical, with no skill in one and not the other. So the field reaches zero on its own as those teams re-sign, which is work already in flight.
- **Deriving it would assert something the original never claimed.** Mapping from the overall verdict or the Tier 1 row would manufacture `failed`/`incomplete` values that the field has never held.

## Verification

With the change, regeneration completes:

```
note: pass_threshold_pct 171 -> 173 nulls — expected while cards migrate; see MIGRATING_FIELDS.
note: validation_status 248 -> 259 nulls — expected while cards migrate; see MIGRATING_FIELDS.
Wrote benchmarks.json: 365 skills
```

`benchmarks.json` is deliberately **not** included here — it is a generated file and the bot regenerates it on the next run once this lands.

Closes #569.


mosheabr:
Admin-merging, with the reason recorded.

**The failing `check` is pre-existing on `main` and is caused by the bug this PR fixes — not by this change.**

The #560 sync changed `nemotron-speech`'s description, so `metadata.json` is stale on main:

```
main metadata.json :  "...Nemotron Speech (Riva) NIM tasks..."
main SKILL.md      :  "...Nemotron Speech (Formerly Riva) NIM tasks..."
```

The regeneration workflow writes `metadata.json` and `benchmarks.json` together. It dies at the `benchmarks.json` null-regression guard before committing either, so `metadata.json` has been stale since 2026-09-15 18:14Z and every hourly run since has failed (#569).

That makes it circular: this PR cannot merge because `check` fails on stale `metadata.json`, and `metadata.json` cannot refresh until this PR merges. Any PR opened right now fails identically, for the same reason.

The alternative — committing regenerated `metadata.json` and `benchmarks.json` here — would put generated files in a human PR, which is exactly what we avoid. So the admin path is the lesser of the two.

This diff is 17 lines of comment plus one set member, touching only `.github/scripts/aggregate_benchmarks.py`. `authors`, `dco`, `content-integrity` and `require-nvskills-ci` all pass. Verified locally that regeneration completes with the change:

```
note: pass_threshold_pct 171 -> 173 nulls — expected while cards migrate; see MIGRATING_FIELDS.
note: validation_status 248 -> 259 nulls — expected while cards migrate; see MIGRATING_FIELDS.
Wrote benchmarks.json: 365 skills
```

Expected after merge: the next hourly run regenerates both files, `check` goes green repo-wide, and the bot opens its own regeneration PR.

PR #574, opened 2026-09-16
Add isaac-mission-control-showcase skill

nv-ppandurangi:
Adds the Isaac Mission Control showcase: an agent skill that runs the Isaac Mission Control tutorial end to end. It detects or installs a local Isaac Sim, launches it with its GUI window, boots a Nova Carter SIL brain, brings up the Mission Control cloud stack, and submits a mission that drives the simulated robot around a small-warehouse scene.

PR #575, opened 2026-09-16
Add isaac-mission-control-showcase skill

mosheabr:
Same-repo copy of #574 so NVSkills CI can sign the skill — signing does not run on fork PRs. @nv-ppandurangi's commit is carried unchanged with its original authorship and DCO sign-off.

Adds the Isaac Mission Control showcase: an agent skill that runs the Isaac Mission Control tutorial end to end. It detects or installs a local Isaac Sim, launches it with its GUI window, boots a Nova Carter SIL brain, brings up the Mission Control cloud stack, and submits a mission that drives the simulated robot around a small-warehouse scene.

### Shape

This is a catalog-native skill — no `components.d/*.yml` entry, like `nvidia-skill-finder`, the `omniverse-*` skills and `physical-ai-neural-reconstruction`. It is authored here rather than mirrored from a source repo, so no sync will move or revert it, and the `isaac-` prefix is carried by the directory and the frontmatter `name` rather than by a `catalog_dir`.

`BENCHMARK.md`, `skill-card.md` and `skill.oms.sig` are intentionally absent — they are generated by the signing run on this PR.

### Review notes

- SPDX headers verified on all 10 code files: `Apache-2.0` (single, not dual), entity `NVIDIA CORPORATION & AFFILIATES.`, placed after the shebang where present.
- `evals/evals.json` is valid JSON with 4 cases.
- `config/skillspector-baseline.yaml` is present ahead of signing, which is required since the baseline is itself a signed resource. It pre-declares 7 suppressions across AE1, PE3 and PE5, each with a specific written rationale; worth confirming which actually fire once Tier 1 has run.

Supersedes #574.


siqinvidia (changes requested):
Thanks for moving this to an in-repo branch and running NVSkills CI. The signing artifacts are now present, content integrity passes, and the benchmark contains real Tier 3 results with an overall PASS.

I found two content issues that need to be resolved before merge:

1. **The documented skill paths do not match the catalog’s flat layout.**

   This PR publishes the skill at:

   `skills/isaac-mission-control-showcase/`

   However, there are total nine references pointing to:

   `skills/isaac-orchestration/isaac-mission-control-showcase/`

   This affects seven executable examples in `SKILL.md`, one expected behavior in `evals/evals.json`, and one reference in `config/skillspector-baseline.yaml`. Commands copied from the current `SKILL.md` will fail because that nested directory does not exist in this PR.

   Please update these references to the actual published path, or use a location-independent mechanism based on the resolved skill directory.

2. **`skill-card.md` contains an unsupported version claim.**

   The generated card states:

   `0.3.0 (source: changelog, released 2025-05-21)`

   I could not find a changelog, a `0.3.0` declaration, or that release date anywhere in the submitted skill bundle. Please correct the source metadata used by the generator so the published card has traceable version information.

After correcting the source content, please rerun `/nvskills-ci` so `skill-card.md`, `BENCHMARK.md`, and `skill.oms.sig` are regenerated and signed against the final files.

siqinvidia (approved):
Re-review complete. The blocking path issue has been fully resolved: all nine references now use the published flat-layout path, and the skill was re-signed afterward.

I also verified that the regenerated benchmark contains real Tier 3 PASS results and that DCO, author verification, content integrity, metadata generation, plugin validation, and NVSkills CI are all green.

The remaining `skill-card.md` version claim (`0.3.0`, sourced from a changelog dated 2025-05-21) is not traceable within this skill bundle. I’m treating this as a non-blocking Skill Card generator follow-up because it does not affect runtime behavior, signature integrity, or evaluation readiness. 

Approved.

